PR #3, opened 2015-01-31
Don't link to regex_macros at runtime.

### LINKED ISSUE #2: A problem with regex_macros
At runtime, I'm getting an error:

`error while loading shared libraries: libregex_macros-[...].so`

after adding a reference to this library.  After asking about this on irc, burntsushi said 

> you need to use `#[no_link] #[plugin] extern crate regex_macros;`

He then added why:

 > the reason why you need all this special stuff is because `regex!` is a compiler plugin and not just a normal macro. in fact, it won't work in the stable channel of Rust 1.0!

Could you fix this or get back to me on how I can continue on without it? Thanks!

BurntSushi:
Fixes #2.

You may also want to consider moving to plain `regex` since `regex_macros` won't work in the stable channel of Rust 1.0. (We may come up with a fix before then using `build.rs` though.)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -169,6 +169,7 @@
 #[macro_use]
 extern crate log;
 #[plugin]
+#[no_link]
 extern crate regex_macros;
 
 extern crate "rustc-serialize" as serialize;


sunng87:
Thanks for the patch! I will move to regex later.

sunng87:
Just an update:

There seems no way to create regexp constant without using `regex!`. So I may need to keep regex_macros until I get a solution.

Issue #18, opened 2015-04-12
RenderError should impl std::error::Error


hugoduncan:
To aid in error conversions and reporting, `RenderError` should implement the `std::error::Error` trait.

PR #20, opened 2015-04-13
Make TemplateError accessible

hugoduncan:
This makes TemplateError available outside of the crate, and also implements
std::error::Error on it.


sunng87:
Thanks for the patch. I will fix the master against nightly then merge this.

(fixes #18)

sunng87:
Thanks for the patch.

PR #21, opened 2015-04-25
use num::FromPrimitive instead

blaenk:
FromPrimitive was removed from the nightly std (not sure about beta), so
we might as well migrate to the num crate which now houses it. This
means we have to manually implement FromPrimitive.

Also no longer need the `core` feature.


sunng87:
Thanks! Seems they have stablizied `FromPrimitive`. We are one more step near to beta-compatible.

PR #22, opened 2015-05-17
compile on stable

blaenk:
This removes the use of the language features to enable its compilation
on rust stable.

the plugin feature was removed and so was the regex_macros dependency.
Instead of re-building the regular expression each time it needs to be
built, I build it once per `Context` instance, which I think is the next
best thing compared to a global, statically initialized regex.

the collections feature was needed for the str::slice_chars function. I
work around this for now by simply ripping out the slice_chars
implementation from the nightly std and embedding it in the code
directly, which we can use at least until the fate of std slice_chars
is determined.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,6 @@ path = "src/lib.rs"
 [dependencies]
 
 regex = "*"
-regex_macros = "*"
 log = "*"
 rustc-serialize = "*"
 num = "*"
--- a/src/context.rs
+++ b/src/context.rs
@@ -4,11 +4,9 @@ use std::collections::VecDeque;
 
 pub struct Context {
     data: Json,
-    default: Json
+    default: Json,
+    array_index_matcher: Regex,
 }
-
-//pub static NULL_VALUE: &'static Json = &Json::Null;
-static ARRAY_INDEX_MATCHER: Regex = regex!(r"\[\d+\]$");
 
 #[inline]
 fn parse_json_visitor<'a>(path_stack: &mut VecDeque<&'a str>, path: &'a String) {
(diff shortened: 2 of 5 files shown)


sunng87:
@blaenk Excellent patch! Thanks for your effort! 

The only thing I think can be better is to move `slice_chars` into a separated module, like `handlebars::support::str`. We can define a trait called `SliceChars` and impl it for `&str`. Then we will still use `source.slice_chars` in template.rs by importing `self::support::str`. Once the API is stabilized, we only need to remove the `use` line.

blaenk:
Alright I'll give that a shot.

blaenk:
This turned out not to work as well as expected. Attempting to use the `slice_chars` method, even if you import the trait, yields the error stating that it's an unstable library feature 'collections'. In order to force the use of the trait's `slice_chars` instead of the inherent one (being triggered via deref coercions) I would have to use UFCS, which makes it even worse than the method I'm using right now since it would be something like:

``` rust
// this worked because deref coercions allows the use of str::slice_chars despite source being String
let mut slice = source.slice_chars(c, min(c+3, source_len)).to_string();

// becomes this. need to explicitly pass an &str, hence &source[..], in order for the trait to match
// alternatively we can impl SliceChars for String and convert there, but the fact remains
// that this is much longer anyways since we have to use UFCS to make it use the trait's
// slice_chars instead of the inherent slice_chars
let mut slice = SliceChars::slice_chars(&source[..], c, min(c+3, source_len)).to_string();
```

Doesn't seem like a clear win to me. I don't think it's too big of an issue to change the 4 locations where we use `slice_chars` this way. It's not entirely clear that it'll eventually be stable anyways, since the unstable message implies that it "has yet to prove its worth." Doesn't seem it merits over-engineering in order to assure a slightly smoother transition.

Let me know what you think.

sunng87:
hmmm... I still hope we can put the `slice_chars` into a support module, based on its functionality. Perhaps we can rename the `slice_chars` in our custom trait to `slice_chars_alt`. WDYT?

blaenk:
Yeah that should work.

On Sunday, May 17, 2015, Ning Sun <notifications@github.com> wrote:

> hmmm... I still hope we can put the slice_chars into a support module,
> based on its functionality. Perhaps we can rename the slice_chars in our
> custom trait to slice_chars_alt. WDYT?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sunng87/handlebars-rust/pull/22#issuecomment-102926484>
> .
>


-- 
- Jorge Israel Peña

sunng87:
Great! Thanks again for your patience!

blaenk:
There you go.

Tests don't build because of the dependency on tojson_macros. I'll leave that to you to sort out.

sunng87:
@blaenk Thanks! Let me deal with that.

PR #25, opened 2015-06-21
Fix example in docs to use data from structure to render.

AlexTalker:
I tried to make an example more useful than it has to be.


sunng87:
Thanks!

PR #27, opened 2015-07-18
Fix @last on a single-element array #each.

chris-morgan:
Formerly, @last would never be true in #each of a single-element array.


sunng87:
Thanks for the fix!

PR #37, opened 2015-09-16
Support mass rendering

### LINKED ISSUE #36: support "mass" rendering
If I have a lot of templates and I want to render all of them with the same context It may be useful to convert the parameter into JSON once and use it on every templates. That would spare a `to_string()` in `renderw()`.

If you are open for this feature I think I can contribute it.

ihrwein:
Fixes #36 

I'd like to write some tests for it before you merge.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "handlebars"
-version = "0.9.3"
+version = "0.9.4"
 authors = ["Ning Sun <sunng@about.me>"]
 description = "Handlebars templating implemented in Rust."
 license = "MIT"
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -95,6 +95,40 @@ impl Registry {
             })
         }
     }
+
+    pub fn render_iter<T>(&self, ctx: &T) -> RendererIterator where T: ToJson {
+        RendererIterator {
+            iter: self.templates.iter(),
+            ctx: Context::wraps(ctx),
+            registry: self
+        }
+    }
+}
+
+pub struct RendererIterator<'a, 'r> {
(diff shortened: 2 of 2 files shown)


sunng87:
Hi @ihrwein , thanks for this patch and your effort. Instead of providing a `RendererIterator`, I think giving user the `get_templates().iter()` is more flexible in general. For example, they can use `map(...)` to render or add a `filter(...)` for selective rendering. The `RenderIterator` seems a little duplicated to template iterator and can be implemented outside the core library.

The issue you pointed out in #36 is some useless `clone()` when calling `render()` multiple times. So I think we should focus on that issue by providing a `render()` with `Json` type as data. 

WDYT?

ihrwein:
You are right, I was too eager to solve this. How would you call the new function? We may also need a new constuctor for `Context` which takes ownership of a `Json` value (If I'm correct).

sunng87:
True, perhaps we need to change `Context` to hold a `&Json` instead of a clone in current impl. 

Also I'm thinking about a break change to `renderw`. We can change `renderw` to accept data as `&Json`. And leave `render()` as a convenience shortcut, unchanged.

sunng87:
Changing `Context` to `Context<'a> { data: &'a Json}` seems expensive. Especially we have function like `extend` that will create new `Context`. 

A simpler approach is to use `Context` as parameter in `renderw`. User could construct a `Context` and reuse it in the iterator.

ihrwein:
That would be fine for me. Should I make a pull request?

sunng87:
@ihrwein Will be great if you would :-) Thanks!

ihrwein:
I sent a new PR. Should I close this one so we can move to the new one with the conversation and review?

sunng87:
Thanks for your patient. I will close this one.

PR #39, opened 2015-09-17
change renderw()'s signature so it takes a Context instead of a T: To…

ihrwein:
…Json

This is a breaking change. To upgrade your code you have to wrap the
original ctx into a Context with Context::wraps() then you can pass a
borrowed reference of it to renderw().

Signed-off-by: Tibor Benke <ihrwein@gmail.com>

### DIFF
--- a/examples/render_file.rs
+++ b/examples/render_file.rs
@@ -69,7 +69,10 @@ fn main() {
     handlebars.register_helper("format", Box::new(format_helper));
 //    handlebars.register_helper("format", Box::new(FORMAT_HELPER));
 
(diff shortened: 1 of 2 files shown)


sunng87:
LGTM. And thanks again for your great patient on solving this issue. 

I will leave it here for several hours to see if there's anyone has concern about it. Then release this as 0.10.0

ihrwein:
OK, thank you very much :smile:

sunng87:
@ihrwein  0.10.0 released!

ihrwein:
Thanks!

Issue #30, opened 2015-07-31
Improve error reporting


sunng87:

gsingh93:
@sunng87 Can you start this off by adding the name of the template that wasn't found to the "Template not found" `RenderError` description?

tailhook:
+1, I don't think the template engine is usable without even tiny bit of error reporting.

sunng87:
I understand. Just need some time for this. Pull requests are always welcomed.

sunng87:
Error report and line_no has been added to 0.11.0 and released to crates.io. I will close this.

PR #41, opened 2015-09-25
Implement some error reporting (#30)

tailhook:
The patch doesn't make error reporting perfect because there are no line numbers yet. But at least this makes error reporting on par with `rumblebars` that have multiple errors types but not error lines too (the `rumblebars` have somewhat complex dependencies, however).

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -1,6 +1,5 @@
 use std::cmp::min;
 use std::ops::BitOr;
-use std::error;
 use std::fmt::{self, Display, Formatter};
 use std::collections::{BTreeMap, VecDeque};
 use std::string::ToString;
@@ -8,6 +7,7 @@ use num::FromPrimitive;
 use regex::Regex;
 
 use support::str::SliceChars;
+use TemplateError;
 
 use self::TemplateElement::{RawString, Expression, HelperExpression,
                             HTMLExpression, HelperBlock, Comment};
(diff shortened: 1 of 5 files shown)


src/template.rs
```diff
@@ -537,7 +522,8 @@ fn test_parse_error() {
 
     let t = Template::compile(source.to_string());
 
-    assert!(t.is_err());
+    assert_eq!(format!("{}", t.unwrap_err()),
+        r#"helper "ifequals" was not closed on the end of file"#);
```
sunng87: just a minor issue causing test failed:

```
---- template::test_parse_error stdout ----

	thread 'template::test_parse_error' panicked at 'assertion failed: `(left == right)` (left: `"helper \"ifequals\" was not closed on end of file"`, right: `"helper \"ifequals\" was not closed on the end of file"`)', src/template.rs:525
```

"on **the** end" vs "on end"

sunng87:
Thanks so much for this patch. I am 100%  ok with this implementation. 

From Oct.1 I will get some holidays to work on this and add line number to the error report.

sunng87:
#30

tailhook:
Fixed typo. Sorry, got spellchecker working after running the test :)

tailhook:
And yes, this also works only on >= Rust 1.2.0 (because the macro system is uglier in previous versions). Also, you should generate docs with Rust >= 1.3.0 (because of a bug in earlier rust too).

I don't think this is a problem, however :)

sunng87:
Never mind. I'm OK to drop support for earlier rust versions (as hyper does). Will merge this.

PR #48, opened 2015-11-09
Additional error info, more generic string arg for RenderError

untitaker:


untitaker:
I tried to use ``{{ else if }}``, which gave me a pretty generic "template not found". `render_error` didn't accept a generic string type, so I refactored it to take `AsRef<str>` instead. I also moved it to the error struct as `new` method.

sunng87:
Thanks! :+1:

PR #49, opened 2015-11-09
Fix bug when iterating through object with empty keys

untitaker:


untitaker:
I didn't yet figure out where the bug is, would apprechiate hints.

sunng87:
Probably the issue is when we try to build a json path with empty string:

```
let new_path = format!("{}/{}/{}", path, param, k);
```

Then we got `path/param/` and failed to get the value from json object. I'm not sure if this is permitted in only implementation.

untitaker:
I believe I have found a bugfix, could you check?

sunng87:
Thanks for the fix. We need think about if removing `""` will cause some other issues for accessing json data structure.

untitaker:
Perhaps you'll have problems if you use `/ident` instead of `./ident`, but I don't really see anything else. Note that I don't really understand the handlebars impl though.

sunng87:
I should have found a better fix and cherry-picked your test case. Thanks for your fix.

sunng87:
I should have found a better fix and cherry-picked your test case. Thanks for your fix.

untitaker:
Yup, that does seem like a better solution. I wonder though what my fix would've broken? I've tried a lot of things and they all continued to work...

On 12 November 2015 16:53:45 CET, Ning Sun <notifications@github.com> wrote:
>I should have found a better fix and cherry-picked your test case.
>Thanks for your fix.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/sunng87/handlebars-rust/pull/49#issuecomment-156146419

-- 
Sent from my phone. Please excuse my brevity.

sunng87:
Your fix should be correct for most common case. But it possibly breaks some incorrect usage like `/a/b` or `/a//b`. I hope we can keep compatibility for these cases so we better choose a safer fix.

PR #51, opened 2015-11-29
Use AsRef wherever possible

untitaker:
We don't need to explicitly copy strings everywhere, passing slices
often suffices, because parsing copies the original string anyway.


sunng87:
Thanks @untitaker !

PR #53, opened 2015-12-16
Feature/serde feature

### LINKED ISSUE #50: serde vs rustc_serialize
I've been told by Nick Cameron and James Miller that the `rustc_serialize` crate is long past expired and that serde should be used instead.

It should probably be setup as feature options (choose either rustc_serialize or serde) so downstream users aren't forced to move over to serde immediately.

If I get a chance I'll give it a try and submit a PR, but don't expect anything for at least a few days

sunng87:
Fixes #50 

Added a feature flag to use serde_json as type system, while keeping rustc_serialize as default implementation. Backward compatible.


sorenhoyer:
Looking forward to seeing this being merged into master :)

sunng87:
This feature is almost down except an example for serde, that works on both stable and nightly channel.

sorenhoyer:
After testing this locally, I now have this new serde_json feature running successfully with my local fork of handlebars-iron, so I can confirm this is working fine :-)

sunng87:
@sorenhoyer Are you on nightly or stable? I'm on nightly and blocked by an upstream (quasi_macros) failure for now.

https://travis-ci.org/sunng87/handlebars-rust/jobs/98979548

sorenhoyer:
Nightly. :-)

sorenhoyer:
For what it's worth I'm getting the same error when doing a "cargo test --features serde_type..." but it still works just fine when just building and using it. I'm still pretty new to Rust so haven't been able to solve this, however this bug only seems to occur when testing (cargo test..) - A similar error were reported here: https://github.com/Byron/google-apis-rs/issues/131 - don't know it that is any help.

sorenhoyer:
https://github.com/sorenhoyer/handlebars-rust/commit/24b9e79e668a459cb81057cdf9332d58a85f9a56 - this seems to solve the quasi_macros compile error, and lets the tests run using "cargo test --features serde_type --no-default-features"

sunng87:
@sorenhoyer thanks! I will try `default-features=false` on serde_codegen

sunng87:
unfortunately I'm getting even more errors with `default-features` turned off on serde_codegen. I'm on nightly 2016-01-03

TimNN:
There have recently been some changes to rustc which broke multiple projects in the @serde-rs organization (see serde-rs/quasi#26 for example).

I am using serde successfully with `nightly-2015-12-28` in another project of mine, so this should work here too.

I've also been experimenting a little bit with this branch and so far it's working great!

sorenhoyer:
@sunng87 just it does work on rustc 1.7.0-nightly (b4707ebca 2015-12-27). If they've made breaking changes once again I guess we'll have to wait a little longer while the rust ecosystem matures a bit :/ What errors did you get?

sunng87:
@TimNN @sorenhoyer thanks for your investigation! I will wait until serde crates are back to normal. 

Interesting that most rust developers are still on nightlies even after half a year of stable release.

sunng87:
We are almost there! I can compile and run the example on nightly 2016-01-03. But it seems the latest nightly breaks `rand` crate again.

sunng87:
Oh I finally get these tests passed! Will merge this soon.

sorenhoyer:
Cool. Great job. Thanks!

sunng87:
This feature has been released in 0.13.0. Let me know if there is any issue with it.

Document is coming soon.

PR #54, opened 2015-12-22
(feat) add support for custom escape functions

TimNN:
This pull request allows the user to override the way output of `{{expr}}` is escaped. 

While my own use case of disabling escaping completely could be achieved by using `{{{expr}}}` and different escape functions could probably be implemented using custom helpers I think this pull request provides a nicer solution.

I especially think that a rust template engine should be as output format agnostic as possible since the environments in which rust is used are more diverse than those in which JS is used.

Rendering performance seems to be unaffected:

```plain
before:
test parse_template  ... bench:     127,948 ns/iter (+/- 12,205)
test render_template ... bench:      18,471 ns/iter (+/- 7,158)

after:
test parse_template  ... bench:     127,924 ns/iter (+/- 18,775)
test render_template ... bench:      18,503 ns/iter (+/- 9,061)
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -91,6 +91,10 @@
 //! }
 //! ```
 //!
+//! #### Escaping
+//!
+//! As per the handlebars spec, output using `{{expression}}` is escaped by default (to be precise, the characters `&"<>` are replaced by their respective html / xml entities). However, since the use cases of a rust template engine are probably a bit more diverse than those of a JavaScript one, this implementation allows the user to supply a custom escape function to be used instead. For more information see the `EscapeFn` type and `Handlebars::register_escape_fn()` method.
+//!
 //! ### Custom Helper
 //!
 //! Handlebars is nothing without helpers. You can also create your own helpers with rust. Helpers in handlebars-rust are custom struct implements the `HelperDef` trait, concretely, the `call` function. For your convenience, most of stateless helpers can be implemented as bare functions.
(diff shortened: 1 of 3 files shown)


sunng87:
Thanks @TimNN , this is a good addition. I will look into this when I got time (My Christmas will be very busy :-))

TimNN:
Take your time :) And if you find any issues, let me know!

sunng87:
@TimNN  I'm OK with this patch. I will merge this. Thanks for this feature!

sunng87:
@TimNN I just find this patch will break handlebars-iron:

```
src/middleware.rs:112:6: 112:21 error: the trait `core::marker::Send` is not implemented for the type `for<'r> core::ops::Fn(&'r str) -> collections::string::String + 'static` [E0277]
src/middleware.rs:112 impl AfterMiddleware for HandlebarsEngine {
                           ^~~~~~~~~~~~~~~
src/middleware.rs:112:6: 112:21 help: run `rustc --explain E0277` to see a detailed explanation
src/middleware.rs:112:6: 112:21 note: `for<'r> core::ops::Fn(&'r str) -> collections::string::String + 'static` cannot be sent between threads safely
src/middleware.rs:112:6: 112:21 note: required because it appears within the type `Box<for<'r> core::ops::Fn(&'r str) -> collections::string::String + 'static>`
src/middleware.rs:112:6: 112:21 note: required because it appears within the type `handlebars::registry::Registry`
src/middleware.rs:112:6: 112:21 note: required because it appears within the type `Box<handlebars::registry::Registry>`
src/middleware.rs:112:6: 112:21 note: required because it appears within the type `middleware::HandlebarsEngine`
src/middleware.rs:112:6: 112:21 note: required by `iron::middleware::AfterMiddleware`
```

TimNN:
Sorry about that, I can look into fixing that later (I'm on mobile right now), but the fix should be as simple as adding `+ Send` (and maybe `+ Sync`) to the type parameter of the Box in the `EscapeFn` type and wherever else required.

sunng87:
Yes, I just added `Send` and `Sync` now it seems work. Thanks for your quick response!

PR #58, opened 2016-01-30
Improving parser performance 

### LINKED ISSUE #56: Performance when compiled under debug
Registering (in debug) a template through register_template_string() which has an embedded 700-line SVG (xml tags) takes something like 3 minutes on my fairly modern computer.  That's orders of magnitude slower than I would expect.

In release mode it happens in less than 2 seconds.

Normally I wouldn't care about debug mode, but my website rebuilds all templates on every page load during debug mode, which is helpful while I'm developing to see changes immediately w/o recompiling.

sunng87:
Fixes #56 

This patch make template parser O(n) for plain text. Comparing to previous implementation, it only peeks string when `{`, `}` or `~` is found.

Also in this implementation, I use [PutBackN](http://bluss.github.io/rust-itertools/doc/itertools/struct.PutBackN.html) iterator to scan the template string, which is better than previous `slice_chars` call.

Thanks for @mikedilger and @TimNN tracked this issue with me.


sunng87:
`cargo bench` on this branch:

```
test parse_template  ... bench:      51,716 ns/iter (+/- 2,886)
test render_template ... bench:      15,905 ns/iter (+/- 395)
```

on master:

```
test parse_template  ... bench:     133,005 ns/iter (+/- 10,715)
test render_template ... bench:      15,602 ns/iter (+/- 2,296)
```

TimNN:
Awesome!

PR #64, opened 2016-02-21
Allow handlebar syntax errors in raw blocks

JoNil:


sunng87:
Thanks, this is better approach for `{{raw}}` indeed.

src/helpers/helper_raw.rs
```diff
@@ -39,4 +13,49 @@ mod test {
         let r = handlebars.render("t0", &());
         assert_eq!(r.ok().unwrap(), "a{{content}}{{else}}hello");
     }
+
```
sunng87: You can remove this file and move these tests to template.rs because `{{raw}}` is no longer a helper.

src/template.rs
```diff
@@ -279,6 +280,7 @@ impl Template {
 
         let mut buffer: String = String::new();
         let mut state = ParserState::Text;
+        let mut old_state = ParserState::Text;
```
sunng87: Do we need to maintain old_state here? Perhaps we can just reset to `ParserState::Text` on exit, like helper.

JoNil: To be honest I couldn't  figure out if we need to track the old state or
not.

I was trying to preserve as much of the parser state as possible for after.

What do you think?

JoNil: I will refactoring the tests

TimNN:
I'm not very familiar with the code, but how will this handle nested `{{#raw}}` elements?

From reading over the code I believe that right now, if the parser encounters two consecutive opening raw tags (ie: `{{#raw}}{{#raw}}`), it will be unable to ever exit the raw state.

I don't know what the desired / expected behaviour is, but I believe that either reporting an error or resuming normal parsing after the first `{{/raw}}` are preferable and easily implemented.

JoNil:
That is true. I willl fix that!

I have no idea what the expected behaviour is for "{{#raw}}{{#raw}}{{/raw}}{{/raw}}".

sunng87:
Oh.. Seems handlebars has a new syntax for raw block in 4.0:

> Raw blocks are available for templates needing to handle unprocessed mustache blocks.
{{{{raw-helper}}}}
  {{bar}}
{{{{/raw-helper}}}}
will execute the helper raw-helper without interpreting the content.
Handlebars.registerHelper('raw-helper', function(options) {
  return options.fn();
});
will render
{{bar}}

http://handlebarsjs.com/block_helpers.html#raw-blocks

Overall, it's a new helper with syntax `{{{{help-name}}}}`, which could access its raw content as text instead of template.

I suggest to implement this new syntax instead of maintaining our own. WDYT, @JoNil ? Let me know if you want to take the task otherwise I will start to think about it ;-)

sunng87:
#65

JoNil:
I don't have nearly enough time to work on it unfortunately :(

sunng87:
@JoNil Thank you all the same, for this great patch that inspired me to improve the raw helper.

 I'm going to close this PR for now. And please let me know if I can help on anything about this library. Thanks!

PR #82, opened 2016-06-12
Pest parser

### LINKED ISSUE #81: Port the template parser Pest
The current custom parser is difficult to maintain, and still contains some issue (for example, Json array literal as parameter). [Pest](https://github.com/dragostis/pest) is a PEG parser looks promising. I will investigate the possibility to port the parser to PEG and Pest.

sunng87:
Fixes #81 

I decided to use [pest](https://github.com/dragostis/pest) to replace current custom parser for maintainability and correctness. 

Tasks:

- [x] Grammar
- [x] Rebuild parser with pest Rdp.queue() results
- [x] Keeping error tracking: report template error with line/column number
- [x] Keeping template element maps, line/column number for each item
- [x] Retain leading whitespaces in text

### DIFF
--- a/src/grammar.rs
+++ b/src/grammar.rs
@@ -34,31 +34,35 @@ impl_rdp! {
 
         param = { literal | reference | subexpression }
         hash = @{ identifier ~ ["="] ~ param }
-        exp_line = _{ name ~ (hash|param)* }
+        exp_line = _{ identifier ~ (hash|param)* }
 
         subexpression = { ["("] ~ name ~ (hash|param)* ~ [")"] }
 
-        whitespace_omitter = { ["~"] }
+        pre_whitespace_omitter = { ["~"] }
+        pro_whitespace_omitter = { ["~"] }
 
-        expression = { ["{{"] ~ whitespace_omitter? ~ exp_line ~
-                                whitespace_omitter? ~ ["}}"] }
+        expression = { ["{{"] ~ pre_whitespace_omitter? ~ name ~
+                                pro_whitespace_omitter? ~ ["}}"] }
 
-        html_expression = { ["{{{"] ~ whitespace_omitter? ~ exp_line ~
-                                      whitespace_omitter? ~ ["}}}"] }
+        html_expression = { ["{{{"] ~ pre_whitespace_omitter? ~ name ~
+                                      pro_whitespace_omitter? ~ ["}}}"] }
+
+        helper_expression = { ["{{"] ~ pre_whitespace_omitter? ~ exp_line ~
+                                       pro_whitespace_omitter? ~ ["}}"] }
 
         invert_tag = { ["{{else}}"]|["{{^}}"] }
-        helper_block_start = { ["{{"] ~ whitespace_omitter? ~ ["#"] ~ exp_line ~
-                                        whitespace_omitter? ~ ["}}"] }
-        helper_block_end = { ["{{"] ~ whitespace_omitter? ~ ["/"] ~ name ~
-                                      whitespace_omitter? ~ ["}}"] }
+        helper_block_start = { ["{{"] ~ pre_whitespace_omitter? ~ ["#"] ~ exp_line ~
+                                        pro_whitespace_omitter? ~ ["}}"] }
+        helper_block_end = { ["{{"] ~ pre_whitespace_omitter? ~ ["/"] ~ name ~
+                                      pro_whitespace_omitter? ~ ["}}"] }
         helper_block = { helper_block_start ~ template ~
                          (invert_tag ~ template)? ~
                          helper_block_end }
 
-        raw_block_start = { ["{{{{"] ~ whitespace_omitter? ~ ["#"] ~ exp_line ~
-                                       whitespace_omitter? ~ ["}}}}"] }
-        raw_block_end = { ["{{{{"] ~ whitespace_omitter? ~ ["/"] ~ name ~
-                                     whitespace_omitter? ~ ["}}}}"] }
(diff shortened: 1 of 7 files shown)


sunng87:
@dragostis FYI

src/grammar.rs
```diff
@@ -0,0 +1,251 @@
...
+// array and object types which may contains whitespace. We will use a real json parser
+// for real json processing
+        literal = _{ string_literal |
+                     array_literal |
+                     object_literal |
+                     number_literal |
+                     null_literal |
+                     boolean_literal }
+
+        null_literal = @{ ["null"] }
```
dragostis: `null_literal` and `boolean_lieral` don't really need to be `atomic`.

src/grammar.rs
```diff
@@ -0,0 +1,251 @@
...
+        boolean_literal = @{ ["true"]|["false"] }
+        number_literal = @{ ["-"]? ~ ['0'..'9']+ ~ ["."]? ~ ['0'..'9']* ~ (["E"] ~ ["-"]? ~ ['0'..'9']+)? }
+        string_literal = { ["\""] ~ (!["\""] ~ (["\\\""] | any))* ~ ["\""] }
```
dragostis: Although it doesn't make any particular difference, I would still recommend making `string_lieral` `atomic` for the sake of consistency.

dragostis:
This is pretty nice overall. 👍

sunng87:
@dragostis Thank you for reviewing this! 

Do you have suggestion for getting line/column number for every matched item? It's very important for debugging and error report. Currently I can manually scan the input string to remember position for every line break. I'm curious about if there is better way to do this with PEG parser.

dragostis:
@sunng87 Yes, `Input` offers this [functionality](http://dragostis.github.io/pest/pest/trait.Input.html#tymethod.line_col). You can use the `Token` `start` and `end` fields in order to see exactly where rules matched.

sunng87:
Perfect! Then there is no roadb

sunng87:
Hi @dragostis , it seems I have finished almost everything except dragostis/pest#60 and one more issue: is it possible to keep leading white spaces in `@{ ( !["{{"] ~ any )+ }`? Because in template we often have meaningful white spaces like indention, I hope we can have some way to keep them.

dragostis:
It's kind of hard to do that without changing the `whitespace` rule not to use them. But then, that would impact performance because you would have to do additional checks extremely often. I can't think of anything simpler than this right now. Maybe adding another type of rule? But that might make everything confusing.

BTW, I've fixed dragostis/pest#60 in dragostis/pest#61.

sunng87:
Thank you for your fast response! I confirm dragostis/pest#61 fixed the line_col issue.

I will leave the whitespace issue for a while to see if there's workaround available.

dragostis:
Maybe one way of fixing the issue with the `whitespace` would be to add a `soi` (start-of-input) rule:

`whitespace = _{ !soi ~ [" "] }`

sunng87:
Thank you! Will try that.

Just benchmarked feature/pest and current master:

feature/pest:
```
test parse_template  ... bench:      23,788 ns/iter (+/- 774)
test render_template ... bench:      13,920 ns/iter (+/- 441)
```

master:
```
test parse_template  ... bench:     460,317 ns/iter (+/- 11,713)
test render_template ... bench:      14,102 ns/iter (+/- 255)
```

We gain huge performance improvement with pest!

dragostis:
Looks awesome!

P.S. Bear in mind that `soi` is not yet implemented, just an idea until we decide what the best approach is.

sunng87:
Just realized that 😂

sunng87:
hi @dragostis just to clarify if `soi` will work for case like `  some text {{value}}\n\tnew line` where the second text token is not "start of input". I wish we can retain `\n\t` in the second text token. This is an important feature when using handlebars to generate plain text like:

```handlebars
{{#each some_list}}
  {{value}}
{{/each}}
```

In current implementation space and line break around `{{value}}` is ignored so there is no way to generate text line by line.

dragostis:
Yes, you can, because you have one big, atomic rule:

```
handlebars = @{ sio ~ stuff }
```

The problem will then be that `stuff` will be atomic too. Hm, we probably need a better solution for this.

dragostis:
Another solution would be to define:

```
whitespace = { [" "]* }
```

Then remove `whitespace` `Token`s from the `Parser::queue_mut()` that you don't need, e.g. between the rules.

Another solution that I'm actually considering is adding some kind of key-value store for rules:

```
whitespace = { !inside ~ ([" "] | ... ) }

block = { ["{{"] ~ toggle(inside) ~ ... ~ ["}}"] ~ toggle(inside) } // OR
block(inside) = { ["{{"] ~ ... ~ ["}}"] }
```

dragostis:
It just occurred to me: you can simply map the queue like this:

```
if token is raw_text:
  if previous token:
    token.start = previous.end
  else:
    toke.start = 0 // raw_text is first token
```

sunng87:
Great idea! It should fix the problem. Let me work on it.

sunng87:
@dragostis The last question, when input mismatches, is it possible to give user a rough position for hint?

sunng87:
Saw the `expected` function on parser, will test it.

sunng87:
All done. Now waiting for 0.3.0 to fix the line_col() and expose input from parser.

dragostis:
I've fixed that issue in `0.3.0` already, but we're having a regression because of lifetimes. I don't really have a good solution for this issue though. There's a single test that isn't passing in [dragostis/pest#61].

PR #86, opened 2016-07-10
Disable rustc-serialize when the serde_type feature is enabled

azerupi:
Fixes sunng87/handlebars-iron#47

You should double check, but I think this should do the trick :)

### DIFF
--- a/examples/partials.rs
+++ b/examples/partials.rs
@@ -1,7 +1,7 @@
 #![allow(unused_imports, dead_code)]
 extern crate env_logger;
 extern crate handlebars;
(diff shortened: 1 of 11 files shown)


azerupi:
Aah sorry, I misspelled a couple of `rustc_ser_type` while testing and it was not caught when I did a find and replace later.

It should be fixed now. Let me know if I should squash the commits.

sunng87:
Thank you @azerupi . It looks good to me.

sunng87:
Released in 0.18.2

PR #88, opened 2016-07-19
Change register file path to AsRef<Path>

### LINKED ISSUE #87: register_template_file() should take the template path as an `AsRef<std::path::Path>`
If register_template_file took its path argument as an `AsRef<std::path::Path>`, then passing in a string would also work.

The new signature would be:

`fn register_template_file<P: AsRef<std::path::Path>>(&mut self, name: &str, tpl_path: P) -> Result<(), TemplateFileError>`

sunng87:
Fixes #87

### DIFF
--- a/examples/render.rs
+++ b/examples/render.rs
@@ -12,7 +12,6 @@ extern crate serde;
 #[cfg(feature = "serde_type")]
 extern crate serde_json;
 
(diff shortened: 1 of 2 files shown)


sunng87:
hmmm seems travis dropped support for rust <= 1.7.0

PR #89, opened 2016-07-22
fixed #each helper to access parent via ../ instead of ../../

metadave:
We're using handlebars-rust in [Habitat](https://www.habitat.sh/). It's fanstastic, thank you!

A [user reported](https://github.com/habitat-sh/habitat/issues/1083) that `../` is not working correctly in an `each` block. Specifically, when a template should be able to use `../` to access a parent value, it instead requires `../../`, as the array index is included in the path as part of the logic in [context:: parse_json_visitor](https://github.com/sunng87/handlebars-rust/blob/565c1353cbc36e81479804510a98123131e0aaf0/src/context.rs#L28). This PR contains a fix for this issue.

Included is a test based off the following manual test via [http://tryhandlebarsjs.com/](http://tryhandlebarsjs.com/):

If I use this template:

```
{{#each a.c}} 
  d={{d}} b={{../a.b}} 
{{/each}}
```

and the following JSON:

```
{
  "a": {
    "b": 99,
    "c": [
      { "d": 100},
      { "d": 200}
    ]
  }
}
```

tryhandlebarsjs.com renders:

```
  d=100 b=99 
  d=200 b=99 
```  

handlebars-rust currently renders the same template and data like this:

```
  d=100 b=
  d=200 b=
```  

Cheers -
Dave

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -32,13 +32,42 @@ fn parse_json_visitor<'a>(path_stack: &mut VecDeque<&'a str>, path: &'a str) {
                 continue;
             }
             ".." => {
-                path_stack.pop_back();
+                let item = path_stack.pop_back();
+                if let Some(item) = item {
+                    if let (Some(a), Some(z)) = (item.chars().nth(0), item.chars().last()) {
+                        // if the item we just popped off is an array index,
+                        // pop off one more to make it behave more like the
+                        // Javascript implementation
+                        if a == '[' && z == ']' {
+                            path_stack.pop_back();
+                        }
+                    }
+                }
             }
             _ => {
                 for dot_p in p.split('.') {
-                    // if dot_p != "this" {
-                    path_stack.push_back(dot_p);
-                    // }
+                    let skip = match (path_stack.front(), path_stack.back()) {
+                        (Some(ref a), Some(ref z)) => {
+                            // if this path starts with "this", then we're not going
+                            // to skip anything
+                            if **a == "this" {
+                                false
+                            } else {
+                                // if the path doesn't start with "this" and
+                                // the last element on the deque is "p",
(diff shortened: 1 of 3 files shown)


metadave:
I see that Travis is unhappy with older versions of Rust (1.4-1.7), I can take a look on Monday.

Have a great weekend!

metadave:
actually, `master` doesn't seem to compile with any of the versions that Travis failed with.

sunng87:
Thank you for using this lib and this pr. Just ignore the travis status, it dropped older rust versions. 

I will look into the patch later today.

sunng87:
Looks good to me! Thank you @metadave .

src/context.rs
```diff
@@ -32,13 +32,42 @@ fn parse_json_visitor<'a>(path_stack: &mut VecDeque<&'a str>, path: &'a str) {
...
+                        (Some(ref a), Some(ref z)) => {
+                            // if this path starts with "this", then we're not going
+                            // to skip anything
+                            if **a == "this" {
+                                false
+                            } else {
+                                // if the path doesn't start with "this" and
+                                // the last element on the deque is "p",
+                                // then don't push it
+                                **z == dot_p
```
sunng87: If the JSON object has same key name in two levels, then there will be no chance to visit the inner one. For example:

```
{
  "a": {
    "a": 99,
    "c": [
      { "d": 100},
      { "d": 200}
    ]
  }
}
```

The second `a` in `../a.a` will be ignored.

sunng87:
hmm..  After some investigation, I think it has something to do with context root with the block.

When using `a.c` with `each`, the context root is at `a`. 

I will try to correct this later.

metadave:
thanks so much @sunng87!

PR #92, opened 2016-07-26
Add example to render arbitrary template from arbitrary json

tailhook:
I think this is very useful to test syntax supported by the library and other things


sunng87:
Great idea, @tailhook ! I think we can just ignore serde backend in this tool. See `render_file` example.

PR #94, opened 2016-08-02
allow identifiers to contain "-"

metadave:
Hello -
  we've found that template expressions that include the `-` character fail to compile. See the included test which fails without this patch.

The Handlebars grammar specifies identifiers quite differently: https://github.com/wycats/handlebars.js/blob/master/src/handlebars.l#L18-L27
but doesn't exclude `-`.

Cheers -
Dave


metadave:
related: https://github.com/sunng87/handlebars-rust/commit/4392580c6c7af41be0866d5f9eb39796fe7330bd

sunng87:
Thank you, @metadave

sunng87:
I'll make a release for this so it won't block your development.

metadave:
thanks so much @sunng87!!

PR #95, opened 2016-08-06
Fix bug where nested scopes don't resolve parent context

jupp0r:
This change fixes a bug where for the template

```
{{#each b}}{{#if ../../a}}{{#each this}}{{this}}{{/each}}{{/if}}{{/each}}
```

and the input

```
{"a": [{"b": true}], "b": [[1, 2, 3],[4, 5]]}
```

the output `123` would be produced. The correct output `12345`
is produced after the fix. While this change definitely fixes the problem,
a cleaner solution would be to have `RenderContext` implement  the `Clone` trait and cloning it in the call to `render` in the `each` helper.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -291,6 +291,18 @@ mod test {
                    "27".to_string());
         assert_eq!(ctx.navigate(".", "this.titles[0]/../age").render(),
                    "27".to_string());
+        assert_eq!(ctx.navigate(".", "titles.[0]/../../age").render(),
+                   "27".to_string());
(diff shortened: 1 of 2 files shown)


sunng87:
Hi @jupp0r , thank you very much for this patch. As you said, a `clone()` function on `RenderContext` will fix this. So I added a `derive` function to `RenderContext` to clone all fields except `writer`.

There is a minor mistake in your template, the correct one is:
`{{#each b}}{{#if ../a}}{{#each this}}{{this}}{{/each}}{{/if}}{{/each}}`

This issue should be fixed in 84fd531 .

Thank you again for this patch.

PR #97, opened 2016-08-14
Fix case where with is used inside an each with nested data access

ghotiphud:
Fix templates like: 

```
{{#each foo}}
    {{#with bar}}
         {{../foovar}}
    {{/with}}
{{/each}}
```

foovar was not being found.


sunng87:
Good fix! Thank you!

Issue #109, opened 2016-10-17
Use a faster HTML escape


Razican:
After the study by @lise-henry in her [article](https://lise-henry.github.io/articles/optimising_strings.html), I think it would be wise to implement it instead of the [current method](https://github.com/sunng87/handlebars-rust/blob/8c0102f9edf7c38e248cc5a335532bb9c9737c21/src/registry.rs#L28). I did some studies myself to try to do it in XML file escaping (not sure if that's for this case, though) and I got to [this version](https://github.com/SUPERAndroidAnalyzer/super/blob/c3e8a723b96743b74297b1798df174549e3fa142/src/results/mod.rs#L813).

sunng87:
@Razican Thank you for this link! Current implementation is pretty straight forward. Would you like to make a pull request for this?

Razican:
Sure!

sunng87:
Implemented in #118

PR #110, opened 2016-10-17
Modified HTML escaping to make it much faster

Razican:
I was able to change the performance of the html escaping as said in #109. But I was unable to use `Cow` due to lifetime issues. This means that when we don't have HTML to scape in the text the function is not as fast as it could be. Here some stats:

```
test bench_hanlebars_old_no_html  ... bench:       1,120 ns/iter (+/- 13)
test bench_hanlebars_old_xml      ... bench:     998,707 ns/iter (+/- 37,883)
test bench_hanlebars_new_no_html  ... bench:         224 ns/iter (+/- 6)
test bench_handlebars_new_xml     ... bench:     316,160 ns/iter (+/- 5,578)
test bench_regex_iter_no_html     ... bench:         198 ns/iter (+/- 4)
test bench_regex_iter_xml         ... bench:     323,962 ns/iter (+/- 4,807)
```

So, in the case of a text with no HTML characters, the escaping is done about 5 times faster (even though it could be a bit faster, from 224ns to 198ns). And in the case of text with HTML (I used a XML file) the reduction is in more that 3 times. In this case seems to be the fastest way to do it.

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -8,6 +8,8 @@ use serialize::json::ToJson;
 #[cfg(feature = "serde_type")]
 use serde::ser::Serialize as ToJson;
 
+use regex::Regex;
+
 use template::Template;
 use render::{Renderable, RenderError, RenderContext};
 use context::Context;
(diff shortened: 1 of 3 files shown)


src/registry.rs
```diff
@@ -25,11 +27,31 @@ pub type EscapeFn = Box<Fn(&str) -> String + Send + Sync>;
...
+                "\"" => output.push_str("&quot;"),
+                _ => unreachable!(),
+            }
+            last_match = end;
+        }
+        output.push_str(&input[last_match..]);
+        output
+    } else {
+        input.to_owned()
+    }
```
Razican: I can test and benchmark it, but it would still mean that we require Regex.

sunng87:
@Razican Thank you for this patch. I decided to use regex `replace_all`, which keeps code clean and easy to maintain. 

Implemented #118

Issue #126, opened 2016-12-22
Extended helper module


sunng87:
I'm going to provide a set of helpers that will be useful to daily development. These helpers will be enabled via a feature called `ext-helpers`. Developers can pick up some of these helpers in their application.

## Number comparison 

- [x] `if-eq`
- [x] `if-neq`
- [x] `if-gt`
- [x] `if-lt`
- [x] `if-gte`
- [x] `if-lte`
- [x] `if-even`
- [x] `if-odd`

## String transform

- [ ] `prefix`
- [ ] `suffix`
- [ ] `join`
- [ ] `uppercase`
- [ ] `lowercase`
- [ ] `capitalize`
- [ ] `substring`

## Array

- [ ] `if-has`
- [ ] `if-not-has`
- [ ] `length`

sunng87:
We will pick some helpers from:
https://github.com/helpers/handlebars-helpers#helpers

killercup:
I tried writing a more generic kind of helper, `gt` (basically `let gt = |x, y| x > y;`), to be used as a subexpression. Having helpers like that would allow you to write much more complex ifs.

Is that currently possibly? How can a helper return a value other than writing to the context?

This is what I have currently:

```rust
fn gt(h: &Helper, _: &Handlebars, rc: &mut RenderContext) -> Result<(), RenderError> {
    let first = h.param(0).map(|x| x.value()).map(|x| x.as_u64()).ok_or_else(|| RenderError::new("`gt` requires two parameters"))?;
    let second = h.param(1).map(|x| x.value()).map(|x| x.as_u64()).ok_or_else(|| RenderError::new("`gt` requires two parameters"))?;
    let result = first > second;

    rc.writer().write(result.into_bytes().as_ref())?;

    Ok(())
}
```

sunng87:
I can understand you want simple and lightweight helper like a javascript function. Handlebars takes subexpression render result as json literal or variable name. To have simple helper like `let gt = |x, y| x > y;`, we just need a wrapper to do two things: 

1. Extract x,y from helper params, check type or report error
2. Write the return value to writer

Perhaps we can wrap it in a macro like `helper!(|x: u64, y: u64| x > y)`, which expands to 

```
|h: &Helper, _: &Handlebars, rc: &mut RenderContext|-> Result<(), RenderError> {
[...]

PR #140, opened 2017-02-28
Helper contrib

sunng87:
More helpers #126

### DIFF
--- a/src/ext.rs
+++ b/src/ext.rs
@@ -1,0 +1,42 @@
+use helpers::HelperDef;
+use registry::Registry;
+// use context::to_json;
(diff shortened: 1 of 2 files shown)


killercup:
Just saw this PR. For a project using handlebars.js, I've added some helpers that allow me to write

```hbs
{{#if (gt 5 x) }}
```

or 

```hbs
{{#if (is-empty x) }}
```

Is this supported by handlebars-rust? If so, it would be a nice alternative to this.

sunng87:
This is possible by defining `gt` and `is-empty` as helper. `()` is syntax for subexpression and inner is parsed and resolve like `{{gt 1 2}}`

PR #164, opened 2017-07-06
bump pest to 0.4

ignatenkobrain:


sunng87:
There's some work to be done in `grammar.rs`. Currently we are waiting for pest 1.0 which is under heavy development.

PR #170, opened 2017-07-31
omit tag escape pattern from resulting RawString element output

softprops:
I think I covered all the cases here. It did seem a bit repetitive but no more repetitive then the existing code for emitting `RawStrings`. 

Ideally I would have liked to find a way to do this in the grammar itself but I couldn't find a clear path for that. I'm new to the pest syntax but familiar with [scala's parser combinator library](https://github.com/scala/scala-parser-combinators#example) and couldn't find the equivalent of scala's `^^` operator to transform the grammar

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -135,6 +135,13 @@ impl Template {
         }
     }
 
+    fn unescape_tags<T>(txt: T) -> String
+    where
+        T: Into<String>,
+    {
+        txt.into().replace("\\{{", "{{")
+    }
+
     fn push_element(&mut self, e: TemplateElement, line: usize, col: usize) {
         self.elements.push(e);
         if let Some(ref mut maps) = self.mapping {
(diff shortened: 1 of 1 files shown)


src/template.rs
```diff
@@ -377,12 +377,18 @@ impl Template {
                     {
                         let (line_no, col_no) = parser.input().line_col(prev_end);
                         if token.rule == Rule::raw_block_end {
-                            let text = &source[prev_end..token.start];
+                            let mut text = &source[prev_end..token.start];
+                            if text.starts_with("\\{{") {
```
sunng87: hi @softprops , what if the `\\{{` is not at beginning of a raw_text, like `hello \\{{world}}`

sunng87: And we can make this a function to avoid duplicated code

softprops: Good call. I wasn't sure of the grammar only parsed raw text with these at the beginning. I updated the tests, impl, and refactored to a method

src/template.rs
```diff
@@ -135,6 +135,13 @@ impl Template {
         }
     }
 
+    fn unescape_tags<T>(txt: T) -> String
+    where
+        T: Into<String>,
```
sunng87: Can we just use `&str` here to avoid creating a `String` just for replace?

softprops: Good call. I didn't see `str` [had that impl](https://doc.rust-lang.org/std/primitive.str.html#method.replace). I'll update and push that change tonight

softprops: fixed

sunng87:
Thank you @softprops for your patient! I will create a release later today.

softprops:
bawler. Ill grab that and pull it into porteurbars so I think I should be set for our company hackathon. Thanks so much for the fast feedback!

Issue #171, opened 2017-08-02
escape tag follow up


softprops:
After integrating the latest tag escaping changes with porteurbars I noticed something. a template in a separate file that contained `\\{{foo}}` was not technically what the current replace `\\{{` with `{{` does. `\` is an escape character in string literals in rust so this was more like replacing `\{{` with `{` leaving behind `\{{foo}}` which wasn't the intended effect. I wanted to report here. I can make a quick follow up pr but I wanted to open an issue for discussion if the pr ends up being the wrong impl so the conversation can live on.

softprops:
I think this [playground link](https://play.rust-lang.org/?gist=6ef530cc02f2b45748c21e303f9ceae2&version=stable) describes it best

sunng87:
Released in 0.28.3

PR #172, opened 2017-08-02
(fix) impl escape that works for displayed strings vs debug string li…

softprops:
fix for #171. tests are updated an change was integration tested against https://github.com/softprops/porteurbars locally

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -396,6 +396,6 @@ mod test {
             "hello": "world"
         });
 
-        assert_eq!("{{hello}}", r.template_render("\\{{hello}}", &data).unwrap());
+        assert_eq!("{{hello}}", r.template_render("\\\\{{hello}}", &data).unwrap());
     }
 }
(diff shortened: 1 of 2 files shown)


src/registry.rs
```diff
@@ -396,6 +396,6 @@ mod test {
             "hello": "world"
         });
 
-        assert_eq!("{{hello}}", r.template_render("\\{{hello}}", &data).unwrap());
+        assert_eq!("{{hello}}", r.template_render(r"\\{{hello}}", &data).unwrap());
```
softprops: so I learned something new here about r(aw) string literals 

```rust
assert_eq!(r"\\{{hello}}", "\\\\{{hello}}")
```

sunng87:
Good catch. I should have noticed that. I will create a fix release later today.

softprops:
Thanks. My bad. I was able to test this directly as a path dependency in a local copy of my crate

PR #188, opened 2017-09-25
(feat) use cow to maintain context-json-lifetime

sunng87:
Currently, in order to bypass lifetime check, JSON data from context are cloned everytime when there is a reference to it. This adds way too much overhead in rendering time. 

This patch is to reduce clones as much as possible by using a `Cow` to hold a borrowed reference (from context) or owned Json (computed)


sunng87:
Close as we have a new branch #196  for this

PR #213, opened 2018-04-13
eliminate use of try!

FauxFaux:
IntelliJ can automatically eliminate `try!`. I have run it.

Afterwards, I ran `cargo +stable fmt`, which printed a load of errors about the rustfmt.toml, but seems to have done roughly the right thing; it's removed some trailing commas in places that they probably weren't needed.


sunng87:
Thanks!

Issue #219, opened 2018-06-14
0.32.4 fails to compile


ignatenkobrain:
```rust
+ cd handlebars-0.32.4
+ /usr/bin/cargo build --release -j6
   Compiling libc v0.2.42
   Compiling serde v1.0.66
   Compiling unicode-xid v0.1.0
   Compiling void v1.0.2
   Compiling quote v0.3.15
   Compiling regex v1.0.0
     Running `/usr/bin/rustc --crate-name libc /usr/share/cargo/registry/libc-0.2.42/src/lib.rs --crate-type lib --emit=dep-info,link -C opt-level=3 --cfg 'feature="use_std"' -C metadata=27b45c2d3f3fe300 -C extra-filename=-27b45c2d3f3fe300 --out-dir /builddir/build/BUILD/handlebars-0.32.4/target/release/deps -L dependency=/builddir/build/BUILD/handlebars-0.32.4/target/release/deps --cap-lints allow -Copt-level=3 -Cdebuginfo=2 -Clink-arg=-Wl,-z,relro,-z,now`
     Running `/usr/bin/rustc --crate-name build_script_build /usr/share/cargo/registry/serde-1.0.66/build.rs --crate-type bin --emit=dep-info,link -C opt-level=3 --cfg 'feature="default"' --cfg 'feature="std"' -C metadata=38421ac63432a745 -C extra-filename=-38421ac63432a745 --out-dir /builddir/build/BUILD/handlebars-0.32.4/target/release/build/serde-38421ac63432a745 -L dependency=/builddir/build/BUILD/handlebars-0.32.4/target/release/deps --cap-lints allow -Copt-level=3 -Cdebuginfo=2 -Clink-arg=-Wl,-z,relro,-z,now`
     Running `/usr/bin/rustc --crate-name quote /usr/share/cargo/registry/quote-0.3.15/src/lib.rs --crate-type lib --emit=dep-info,link -C opt-level=3 -C metadata=fe9d3185ffca1f80 -C extra-filename=-fe9d3185ffca1f80 --out-dir /builddir/build/BUILD/handlebars-0.32.4/target/release/deps -L dependency=/builddir/build/BUILD/handlebars-0.32.4/target/release/deps --cap-lints allow -Copt-level=3 -Cdebuginfo=2 -Clink-arg=-Wl,-z,relro,-z,now`
[...]

PR #220, opened 2018-06-14
bump pest_derive to 1.0.7

ignatenkobrain:
Fixes: https://github.com/sunng87/handlebars-rust/issues/219
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>


sunng87:
Thanks. Normally cargo update should fix this. But I'm OK with it explicitly.

ignatenkobrain:
@sunng87 no, the problem is that if you have 1.0.6, then your code won't work. With 1.0.7 works fine. Which means you have to be explicit that you need >= 1.0.7

PR #226, opened 2018-07-12
(feat) added `Registry.has_template` utility method

dathinab:


dathinab:
I also agree to this code pice being re-licensed to MIT/Apache-v2 dual license at any point in time.

src/registry.rs
```diff
@@ -268,6 +268,11 @@ impl Registry {
         &*self.escape_fn
     }
 
+    /// Return `true` if a template is registered for the given name
+    pub fn has_template(&self, name: &str) -> bool {
+        self.get_template(name).is_some()
```
sunng87: why not use `contains_key` on `self.templates`?

dathinab: Just something I got used, too.

When I have the choice to access some data through the (semi) stable
public methods or the internal details and accessing it through the
public methods doesn't introduce extra cost I always use the public
methods.

This has the benefit that the thing I wrote just depends on the public
interface and not the implementation details, so if I refactor the implementation
I don't have to change/update the method, at last as long as the public
interface part I used did not change. Also if there is some more handling
added to `get_template` this would also implicitly apply to `has_template`.
(i.e. putting a brake point on `get_template`)

It is similar to how futures 0.2 splits the future trait into the trait and
`FutureExt`, just that for futures there are some really good reasons
to do so while here it's mostly because I'm used to doing things this
way.

For clarification:  The reason I say it doens't have a cost is, because
the compiler should inline this. In the past I always added `#[inline]`
annotations, but then I often couldn't decided wether or not to add them
on some slightly more complex methods braking the flow, so I decided
to stop doing so as llvm/rustc really should be able to do reasonable
good in-lining for higher optimization levels.

sunng87: Thanks for detailed explanation.

PR #228, opened 2018-07-15
Fix a few typos in the docs

killercup:
All credit goes to https://github.com/Jason-Rev/vscode-spell-checker


sunng87:
Seems I should enable flyspell in emacs by default :laughing:

PR #231, opened 2018-08-27
update some debs for minimal-versions

Eh2406:
This bumps the minimal acceptable versions in Cargo.toml to versions that are compatible with `-Z minimal-versions`. This is part of the process of seeing how hard this is for crates to use in preparation for getting it stabilized for use in CI, specifically upstreaming the changes required to get criterion working with it. It is easy to use if all of your dependencies support it, but much harder if trying to impose it on them.

Also cargo defaults to interpreting version requirements as `^`. So this switched to the shorter form.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,17 +22,17 @@ path = "src/lib.rs"
 log = "0.4.0"
 quick-error = "1.0.0"
 pest = "1.0.0"
-pest_derive = "1.0.7"
+pest_derive = "1.0.8"
 serde = "1.0.0"
 serde_json = "1.0.0"
-regex = "1.0.0"
+regex = "1.0.3"
 lazy_static = "1.0.0"
-walkdir = { version = "2.1.4", optional = true }
+walkdir = { version = "2.2.3", optional = true }
 
 [dev-dependencies]
-env_logger = "0.5.0"
+env_logger = "0.5.13"
 maplit = "1.0.0"
-serde_derive = "1.0.0"
+serde_derive = "1.0.75"
 tempfile = "3.0.0"
 
 [features]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,21 +19,21 @@ path = "src/lib.rs"
 
 [dependencies]
 
-log = "^0.4.0"
-quick-error = "^1.0.0"
-pest = "^1.0.0"
-pest_derive = "^1.0.7"
-serde = "^1.0.0"
-serde_json = "^1.0.0"
-regex = "^1.0.0"
-lazy_static = "^1.0.0"
-walkdir = { version = "^2.1.4", optional = true }
+log = "0.4.0"
+quick-error = "1.0.0"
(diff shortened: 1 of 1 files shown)


sunng87:
Thanks! I will create a release later today.

Eh2406:
So there was a [small hiccup](https://github.com/japaric/criterion.rs/pull/196), criterion has a policy of only using crates that build on Rust 1.20. Witch handlebars 1.0 does not (do to pest 1.0). Would you except a PR for the 0.32 branch, or have you moved on to much for it to be worth my time?

sunng87:
But the 0.32 branch [depends on pest 1.0](https://github.com/sunng87/handlebars-rust/blob/0.32.4/Cargo.toml) too. And I am not going to maintain versions prior to 0.32

Eh2406:
Cc @bheisler can you give a little more detail on the problems with jumping to handlebars 1.0. My interpretation of your summary appears to be incorrect.

sunng87:
I see. It was new features introduced in rust 1.28 iirc. I will fix and release a newer version.

Eh2406:
I have made 11 PRs two 10 different projects in the past 2 weeks to get `minimal-version` support for Criterion. It may be selfish of me but I'd like to see that land. If it has to wait so be it.

But, If one more PR can make it happen I can try. One option is to attempt to get `minimal-version`s working for the `0.32` branch. Another is to get `1.0` to work on Rust 1.23.0; All of the errors appear to be about "match ergonomics" so may be easily fixable. And the last option is for me to learn to be more patient.

sunng87:
@Eh2406 If we can get 1.0 work with Rust 1.23, do we still need to port it to 0.32? I think we can update handlebars to 1.0 in criterion and get minimal-version work. 

I'm 100% willing to help handlebars 1.0 to land on criterion.

Eh2406:
> I will fix and release a newer version.

Thank you. And thank you for all the work you put into handlebars. If I can help let me know.

> do we still need to port it to 0.32

My reading is no, but I'd like @bheisler to confirm that I understood correctly before we put in the work.> I will fix and release a newer version.

Thank you. And thank you for all the work you put into handlebars. If I can help let me know.

> do we still need to port it to 0.32

My reading is no, but I'd like @bheisler to confirm that I understood correctly before we put in the work.

bheisler:
If you can get handlebars to compile on 1.23.0, I think I could update without any other problems.

>I have made 11 PRs two 10 different projects in the past 2 weeks to get minimal-version support for Criterion. It may be selfish of me but I'd like to see that land. If it has to wait so be it.

Thank you for all of your work as well - I didn't expect you to submit pull requests to all of my dependencies just for this. You've really gone above-and-beyond on this.

sunng87:
@Eh2406 @bheisler just released 1.0.3 and it should build on rust 1.23.0. I have setup travis build to ensure future development works with 1.23

Eh2406:
Thank you for making that work!

PR #235, opened 2018-09-30
(chore) fix the url in #![doc(html_root_url)]

ubnt-intrepid:
See the Rust API guideline ([link][1]) for details.

[1]: https://rust-lang-nursery.github.io/api-guidelines/documentation.html#crate-sets-html_root_url-attribute-c-html-root

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,4 @@
-#![doc(html_root_url = "https://docs.rs/crate/handlebars/")]
+#![doc(html_root_url = "https://docs.rs/handlebars/1.0.5-alpha.0")]
 //! # Handlebars
 //!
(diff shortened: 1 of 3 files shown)


sunng87:
Thank you for this patch. For adding version number to the line, can you please use cargo-release's replacement hook? You can see release.toml for detail.

ubnt-intrepid:
@sunng87 added entries to replacement hook.

src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![doc(html_root_url = "https://docs.rs/crate/handlebars/")]
+#![doc(html_root_url = "https://docs.rs/handlebars/CrateVersion")]
```
sunng87: There is an issue if you hard-coded `CrateVersion` here that once it's replaced in a release. There will be no such `CrateVersion` to replace in future release. You can use regex to capture the version number here.

There is an example in cargo-release's release.toml as example. Replying from mobile, sorry for unable to link it.

ubnt-intrepid: I've fixed in e6de93a.

Issue #179, opened 2017-08-28
Segfault in Handlebars


matthiasbeyer:
I just experienced a segfault which seems to be originated in the handlebars code.

As there are ~6000 stack levels, it seems that I ended up crashing into the heap by recursing too deep. The recursion seems to happen in handlebars, as only a few (as in below 10) stack levels are from my code.

The corrosponding code [is here](https://github.com/matthiasbeyer/imag/pull/1023), [here is the causing call](https://github.com/matthiasbeyer/imag/pull/1023/files#diff-63adbd22576587d2d4651b4e1f25c83aR151).

The templates can be found [here](https://github.com/matthiasbeyer/imag/pull/1023/files#diff-bc9ab60707442f7ad22f3b3b34c1da72), the registration of [the templates and the helpers is here](https://github.com/matthiasbeyer/imag/pull/1023/files#diff-63adbd22576587d2d4651b4e1f25c83aR82). If you have any questions regarding the code, feel free to ask me, I know the codebase is pretty big and maybe confusing for some people.

---

To reproduce:

* Clone [the repository](https://github.com/matthiasbeyer/imag), checkout the branch from the PR
* `cargo build --manifest-path bin/core/imag-store/Cargo.toml` # takes some time
* `mkdir /tmp/store/`
* `./target/debug/imag-store --rtp /tmp/ --config imagrc.toml create -p test entry -c foo`
* enjoy.

The last few lines of the stacktrace follow

---

```
#6274 0x00005555556808c2 in core::result::Result<&handlebars::template::Template, handlebars::error::RenderError>::and_then<&handlebars::template::Template,handlebars::error::RenderError,(),closure> (self=Ok = {...}, op=closure = {...})
    at /build/rustc-1.17.0-src/src/libcore/result.rs:601
#6275 0x00005555555f8b18 in handlebars::registry::Registry::renderw<collections::btree::map::BTreeMap<&str, collections::string::String>> (self=0x7ffff6ab9038, name="DEBUG", data=0x7fffffff5578, writer=&mut Write)
    at /home/m/.cargo/registry/src/github.com-1ecc6299db9ec823/handlebars-0.29.0/src/registry.rs:247
#6276 0x00005555555f8654 in handlebars::registry::Registry::render<collections::btree::map::BTreeMap<&str, collections::string::String>> (self=0x7ffff6ab9038, name="DEBUG", data=0x7fffffff5578)
[...]

PR #236, opened 2018-09-30
Add disable compiletime feature to disable logging

matthiasbeyer:
If one uses a logger implementation which uses handlebars for "styling"
the output, logging calls in handlebars itself would start a recursion
and ultimatively crash the program by recursing indefinitly.

If the feature "logging" is not compiled into the library, the "log"
crate is ignored and the macros result in all logging calls to be
ignored.

By default, the "logging" feature is enabled to keep backwards
consistency.

Signed-off-by: Matthias Beyer <mail@beyermatthias.de>

---

Hi,

I have an issue related to the debug output handlebars writes. The Problem is that I use handlebars in my logger implementation, but as handlebars itself logs via that logger of course, I get a recursive execution of the logger.

Because of this I had to add a [nasty hack in my codebase](http://git.imag-pim.org/imag/tree/lib/core/libimagrt/src/logger.rs#n133) to prevent handlebars from calling the logger. This is, of course, not a nice thing to have in a codebase.

This patch adds a feature which would solve my issue. Please consider including it in handlebars-rust.


sunng87:
Thanks! This will permanently fix #179. Could you please check the failing test?

matthiasbeyer:
My bad, closed a parentheses too often. If you're fine with the patchset I'll squash-rebase! :smile:

matthiasbeyer:
Awesome, thanks.

I hope there will be a new release soon, so I can remove my workaround. No hurries though!

sunng87:
Thanks. I'm on vocation these days and having limited internet access. Could you please use git dep at the moment? I will notify you in this thread when new release is ready.

matthiasbeyer:
Awesome! Have a nice vacation! <3

sunng87:
Released in 1.0.5
I added a `no_logging` feature to disable log

PR #259, opened 2019-04-07
Stripping newlines for helper tag

### LINKED ISSUE #258: stripping lines that are "standalone" helpers
In the handlebars.js docs the following is described (https://handlebarsjs.com/expressions.html):
```
This expands the default behavior of stripping lines that are "standalone" helpers (only a block helper, comment, or partial and whitespace).
```
They give the example:
```
{{#each nav}}
  <a href="{{url}}">
    {{#if test}}
      {{title}}
    {{^}}
      Empty
    {{/if}}
  </a>
{{~/each}}
```
That will render as
```
<a href="foo">
    bar
</a>
<a href="bar">
    Empty
</a>
```

Thus rendering the line `{{#if test}}` as nothing. Currenly on handlebars.rs:
```
----------
{{#if body}}
{{ body }}
{{/if}}
```
renders as
```
----------

body here
```
(notice the newline between '---' and 'body here'.) Is it possible to ignore those lines?

sunng87:
Fixes #258

### DIFF
--- a/tests/strip-empty-lines.rs
+++ b/tests/strip-empty-lines.rs
@@ -1,0 +1,24 @@
+extern crate handlebars;
+
+#[macro_use]
+extern crate serde_json;
+
+use handlebars::Handlebars;
+
+#[test]
+fn test_stripping_empty_line_258() {
+    let hbs = Handlebars::new();
+
+    let data = json!({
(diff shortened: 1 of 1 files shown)


sunng87:
This is finally implemented in #404

PR #263, opened 2019-04-26
Fix partial context

### LINKED ISSUE #260: Block parameters can't be used as context for partials
This test case fails:

```rust
#[test]
fn test_partial_with_blocks() {
    let hbs = Handlebars::new();

    let data = json!({
        "a": [
            {"b": 1},
            {"b": 2},
        ],
    });

    let template = "{{#*inline \"test\"}}{{b}};{{/inline}}{{#each a as |z|}}{{> test z}}{{/each}}";
    assert_eq!(
        hbs.render_template(template, &data).unwrap(),
        "1;2;"
    );
}
```

I've verified on http://tryhandlebarsjs.com/ that this is the right output.

sunng87:
Fixes #260 

Fix partial context to use standalone JSON value instead of reference path.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ quick-error = "1.0.0"
 pest = "2.1.0"
 pest_derive = "2.1.0"
 serde = "1.0.0"
-serde_json = "1.0.0"
+serde_json = "1.0.39"
 regex = "1.0.3"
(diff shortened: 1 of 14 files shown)


sunng87:
Finally got #260 and #264 fixed in this path.

Block parameter is completed rewritten to support path access on them.

PR #265, opened 2019-04-28
Swap order of block parameters in each helper

thegedge:
As per https://handlebarsjs.com/builtin_helpers.html#iteration, the order of block parameters is `value` first, `key` second.


sunng87:
Good catch! Thanks!

PR #266, opened 2019-05-07
helpers: accept any json value for boolean helpers

Keruspe:
Makes thinkgs more consistent with default if behaviour, allows writing

```
{{#if (or a b)}}FOO{{else}}BAR{{/if}}
```

instead of

```
{{#if a}}FOO{{else}}{{#if b}}FOO{{else}}BAR{{/if}}{{/if}}
```

when a or b can be null, for example


sunng87:
Great! Thanks!

PR #272, opened 2019-06-24
export BlockParams

Keruspe:
Otherwise we can't use `local_rc.push_block_context` in custom tags

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -351,7 +351,7 @@ extern crate walkdir;
 
 extern crate hashbrown;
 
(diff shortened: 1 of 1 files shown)


sunng87:
That was fast! Thanks! May I see your custom helper? 

For now helper development experience isn't a priority but I'm curious about your usage.

Keruspe:
Here it is:
https://github.com/sozu-proxy/amq-protocol/blob/master/codegen/src/templating.rs#L156

PR #273, opened 2019-06-27
(feat) attempt to add support for string index access

sunng87:


sunng87:
At the moment, Serde doesn't support **borrowed** index access on `String`. This patch won't work.

PR #276, opened 2019-07-07
Support singular and plural block params for each helper

thegedge:
With a single block param, `each` should set it as the value of the container being iterated over.

With two block param, `each` should set the first as the value of the container being iterated over, and the second as the key/index.


sunng87:
Thanks!

PR #280, opened 2019-07-18
(feat) switch back to std hashmap [#279]

### LINKED ISSUE #279: Issues with hashbrown 0.4 -> 0.5 upgrade
`hashbrown::HashMap` is part of the public API of this crate in several places, e.g.:

https://docs.rs/handlebars/2.0.1/handlebars/struct.Handlebars.html#method.get_templates

Between Handlebars 2.0.0 and 2.0.1, `hashbrown` was updated from 0.4 to 0.5. This broke installation of my crate (`cargo install abscissa`), since it was using `hashbrown` v0.4.

Some options I can think of:

- Re-export either `hashbrown` (e.g. `pub use hashbrown` in the crate root) or `hashbrown::HashMap` so I don't have to depend on the `hashbrown` crate
- Wrap `hashbrown::HashMap` in a newtype
- Switch to `std::collections::HashMap` as `std` is switching to `hashbrown` anyway

sunng87:
Fixes #279

Benchmark shows std hashmap now shares same performance with hashbrown. So we'd better swtich back in next major release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,6 @@ serde_json = "1.0.39"
 regex = "1.0.3"
 lazy_static = "1.0.0"
 walkdir = { version = "2.2.3", optional = true }
-hashbrown = { version = "0.5.0", features = ["serde"] }
 
 [dev-dependencies]
 env_logger = "0.6.0"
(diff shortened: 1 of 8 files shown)


sondr3:
FYI, `hashbrown` has become the `std::hashmap` implementation: rust-lang/rust#58623. This is probably why the performance is identical :smile:

PR #282, opened 2019-07-20
Add support for helper contains non-static borrowed data

### LINKED ISSUE #281: Attach custom data to Helper.
Hi!

I'm new to both rust and handlebar-rust, so I'm not quite sure if this is even possible.
Say I want to create a helper that references something else:

```rust
pub struct FooHelper<'a> {
    foo: &'a Foo,
}

let foo = Foo{}
handlebars.register_helper("foo_helper", Box::new(FooHelper{foo: &foo}));
```

This won't work as `register_helper` needs the helper to have `'static` lifetime. Does this mean that I can not attach a reference to the helper? Or maybe this is how it is intended to avoid some behaviors?

Thanks in advance!

sunng87:
Fixes #281

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -57,16 +57,16 @@ pub fn no_escape(data: &str) -> String {
 /// The single entry point of your Handlebars templates
 ///
 /// It maintains compiled templates and registered helpers.
-pub struct Registry {
+pub struct Registry<'reg> {
     templates: HashMap<String, Template>,
-    helpers: HashMap<String, Box<dyn HelperDef + 'static>>,
-    directives: HashMap<String, Box<dyn DirectiveDef + 'static>>,
+    helpers: HashMap<String, Box<dyn HelperDef + 'reg>>,
+    directives: HashMap<String, Box<dyn DirectiveDef + 'reg>>,
     escape_fn: EscapeFn,
     source_map: bool,
     strict_mode: bool,
 }
 
-impl Debug for Registry {
+impl<'reg> Debug for Registry<'reg> {
     fn fmt(&self, f: &mut Formatter) -> Result<(), fmt::Error> {
         f.debug_struct("Handlebars")
             .field("templates", &self.templates)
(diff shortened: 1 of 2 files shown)


jagt:
Hi!

Thanks for you quick feedback.
I've just tried this change out and noticed that maybe it won't work.

1. It's too much of a change as now `Registry` needs a lifetime. I think it's too much of a change that breaks too many other user's code and maybe not really worth it.
2. My use case is that I put a `Registry` into another parent struct, and in this case I just can't get the lifetime to work. It's something like this:

    ```rust
    struct Parent<'a> {
        pub handlebars: Handlebars<'a>,
    }

    pub struct ParentHelper<'a> {
        parent: &'a Parent,
    }

    let foo = Foo{}
    parent.handlebars.register_helper("parent_helper", Box::new(ParentHelper{parent: &parent}));
    ```
   
   I still can't get this to work.

I'll stick with using helpers as pure functions and it would still work for me.
Again thanks for you works!

lfairy:
@jagt Looks like you've got a circular reference. `Parent` owns `Handlebars` which owns `ParentHelper`, but `ParentHelper` references `Parent`. The easiest solution would be to refactor your code to remove the circular dependency.

sunng87:
@jagt you don't have to hold `Handlebars` in the struct, you can get it as argument from the `call` function of helper.

sunng87:
This change works for previous code because all previous tests passed. Actually I was curious about why rust compiler didn't require lifetime parameter for this case.

jagt:
@lfairy yeah I now understand I'm having a self referencing struct and trying to get rid of it. 

@sunng87 I'm pretty sure it would break some code, see the comparison below.

```rust
// before
struct Parent {
    pub handlebars,
}

// after this pr
struct Parent<'a> {
    pub handlebars: Handlebars<'a>,
}
```
Unfortunately I'm so new to rust that I don't really know how this works either...

Cheers!

sunng87:
I see. When handlebars is captured in a struct now it requires developer to specify lifetime for the struct. In out test cases, it's all used as local variables. 

I'll leave this for 3.0 release.

Issue #294, opened 2019-12-08
Addressing performance issue in large loop


sunng87:
As from https://www.vincentprouillet.com/blog/tera-v1-is-here/ handlebars had an unreasonable performance loss in large loop. 

Before looking into very detail of this, there are some know reason may affect the perf:

* Json path are parsed several times when accessing data
* ~~Local variable like `@index` can be generated lazily~~
* Parameter can be parsed during template parsing instead of render time

sunng87:
With #295 and #298 we are now significantly faster than 3.0.0-beta.1

![image](https://user-images.githubusercontent.com/221942/71662387-4be21300-2d8c-11ea-810c-ad3e9815e09b.png)

sunng87:
As of #307 , we have performance improvement for nearly 5x to 2.x. I'm closing this for now.

PR #295, opened 2019-12-25
Path cache

sunng87:
This patch refactored path structures in RenderContext and removed a few duplicated path parsing. First part for #294 

My local tests shows it improves the large loop benchmark for 20%.


sunng87:
Criterion benchmark:

* render_template  -37% 
* large_loop_helper -27.5%
* large_nested_loop -19%

sunng87:
With the render context optimization in each helper, comparing to master we are:

* render_template -42.6%
* large_loop_helper -40.4%
* large_nested_loop -47.8%

PR #309, opened 2020-01-21
(feat) use pprof-rs for profiling

sunng87:


sunng87:
Pending for: https://github.com/tikv/pprof-rs/pull/10

PR #315, opened 2020-02-02
Document register_templates_directory being dependent on the dir_source feature

rukai:
I had to check the source code to figure out what was going on, this should help others avoid needing to do that.


sunng87:
Thank you!

PR #319, opened 2020-03-06
removed `Send + Sync` bound of `RenderContext.register_local_*`

HellButcher:
* removed `Send + Sync` bound of `HelperDef` for `RenderContext::register_local_helper` and related functions.
* removed `Send + Sync` bound of `DecoratorDef` for `RenderContext::register_local_decorator` and related functions.
* refined some `'static` lifetimes to `'rc`

Background:
I use the `RenderContext` inside a HTTP request (the `Registry` is shared), and I want to register
local-helpers that access the HTTP request (for example a helper that resolves a URL)

### DIFF
--- a/src/decorators/mod.rs
+++ b/src/decorators/mod.rs
@@ -56,7 +56,7 @@ pub type DecoratorResult = Result<(), RenderError>;
 /// }
 /// ```
 ///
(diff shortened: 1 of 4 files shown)


sunng87:
LGTM. Thank you! I wonder if there will be break change to existed users.

PR #322, opened 2020-04-11
Fix incorrect parse error for partial identifiers

### LINKED ISSUE #321: Using path as partial block identifier causes incorrect parse error
I believe I have encountered an incorrect parse error when using paths as identifiers for partial blocks.

The use case that brought this up was with the `Handlebars::register_templates_directory` which registers templates and partials with the name set to the relative path from the root of the templates directory.

```txt
- templates-dir/
  - bar.hbs
  - a/
    - foo.hbs
```

Here we have the partial `foo.hbs`. When we setup the `Handlebars` registry with `Handlebars::register_templates_directory` we can then refer to `foo.hbs` as `a/foo.hbs`.

The problem arises when we have:

```hbs
{{! foo.hbs}}
{{#> a/bar}}
  {{#*inline "content"}}{{/inline}}
{{/a/bar}}
```

```hbs
{{! bar.hbs}}
{{#> content}}{{/content}}
```

Specifically, the parse error shows up on the closing expression of the partial block (the line `{{/a/bar}}` in `foo.hbs`.

This example could be reduced just to the core problem, being the parse error of `{{#> a/bar}}{{/a/bar}}` but I thought the example would give more of a hint as to the use case for this.

The error can be checked/reproduced with the test added in #322

joshleeb:
This PR fixes a parse error that arises when using paths, containing `/`, as partial block identifiers.

Closes #321

### DIFF
--- a/src/grammar.pest
+++ b/src/grammar.pest
@@ -71,7 +71,7 @@ decorator_block = _{ decorator_block_start ~ template ~
 
 partial_block_start = { "{{" ~ pre_whitespace_omitter? ~ "#" ~ ">"
                         ~ partial_exp_line ~ pro_whitespace_omitter? ~ "}}" }
-partial_block_end = { "{{" ~ pre_whitespace_omitter? ~ "/" ~ identifier ~
+partial_block_end = { "{{" ~ pre_whitespace_omitter? ~ "/" ~ partial_identifier ~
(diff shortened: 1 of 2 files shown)


sunng87:
Good catch! Thank you!

sunng87:
Released in 3.1.0-beta.2

PR #331, opened 2020-06-06
Add ability to register already-parsed templates

mkantor:
I have a use case for doing template validation separately from registry setup. With this API I could avoid parsing the templates twice.


sunng87:
Sounds good to me. 

The lint error was due to an update of clippy rule. I'll merge and fix later.

mkantor:
Thanks!

PR #290, opened 2019-09-22
Improve wording / grammar in the README

banool:
I tried to not change the meaning of any of the README, I just wanted to make it read a bit better 😃.

PR #333, opened 2020-06-06
Fix some typos/grammar in README and doc strings

mkantor:
I was reading through the docs and noticed a few typos, so fixed them as I went along.


src/lib.rs
```diff
@@ -44,17 +44,17 @@
 //!
```
mkantor, line 44: I copied this back from the README (it was updated in #290).

sunng87:
Thank you for this patch!

PR #338, opened 2020-06-10
Hash support for helper macro

### LINKED ISSUE #335: Support for named args in `handlebars_helper` macro?
It seems it's currently not possible to use the `handlebars_helper` macro for defining helpers with optional named args.
Could support be added by using `hash_get`? :)

The syntax for these args could be `arg: Option<Type>` or `arg=Option<Type>`.

sunng87:
Fixes #335 

This patch added support for named argument in helper macro `handlebars_helper!`.  The syntax is like:

```rust
handlebars_helper!(is_above, |x: u64, { compare: u64 = 10 }| x > compare);
```

Multiple named arguments are supported with `,` separated. 

When using this helper, you can either 

```handlebars
{{is_above 100 compare=99}} or {{is_above 100}}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased](https://github.com/sunng87/handlebars-rust/compare/3.1.0...Unreleased) - ReleaseDate
 
 * [Added] API to register an pre-processed template [#331]
+* [Added] Helper macro now has support for named argument and helepr hash [#338]
 * [Changed] Update rhai to 0.15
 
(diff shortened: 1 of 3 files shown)


Boscop:
Thanks :) Although for many of my use cases it would be more useful to not require a default value / to allow `None` as default value. 
(Not all args are supposed to have values, and the helper body needs to know which ones don't have values.)

sunng87:
The idea of default value is borrowed from python. But, as you said, dealing with None/null was a little different from in python or javascript. Let me think if there is a way to make it consistent while keep it simple.

PR #341, opened 2020-06-16
Helper macro enhancement

sunng87:
Added support for python's `*args` and `**kwargs` style arguments support for helper macro. 

* `*args` provides a vector of all helper parameters as `&serde_json::Value`
* `**kwargs` provides a map of all helper hash `<&str, &serde_json::Value>`

A usage example is included in test case.


sunng87:
@Boscop

PR #355, opened 2020-07-12
Derived value support for with helper 

### LINKED ISSUE #353: With and lookup subexpression
Thanks for this handlebars implementation I have been enjoying using it very much. I think I *might* have hit a bug...

Not sure if this is related to #343 but it appears that there is a bug using lookup in a sub-expression in conjunction with a with block helper.

This code does not work:

```handlebars
{{#with (lookup context.config.authors "tmpfs")}}
{{json this}}
{{/with}}
```

Note that `json` is just a helper that serializes and prints as JSON for  debugging. In this case the context is not changed and `this` refers incorrectly to the outer context.

But I know the object exists because if I change it to:

```handlebars
{{#with context.config.authors.tmpfs)}}
{{json this}}
{{/with}}
```

Then the context is correct.

I am trying to do dynamic lookup of a key in an object so this is all actually going to be a partial called like this `{{>author id="tmpfs"}}` but at the moment it looks like I can't achieve what I want.

I took a look at the code and I thought at first it was an issue with the `lookup` helper but adding a test case to the code made me think it may be an issue with the `with` helper instead.

I will put together something simple to try to reproduce it shortly then maybe you can advise where the issue might be. Thanks!

sunng87:
Fixes #353

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## [Unreleased](https://github.com/sunng87/handlebars-rust/compare/3.3.0...Unreleased) - ReleaseDate
+
+* [Fixed] Derived value as block context base value [#343, #353]
+
 ## [3.3.0](https://github.com/sunng87/handlebars-rust/compare/3.2.1...3.3.0) - 2020-07-18
 
 * [Added] Added two new APIs to reuse `Context` for rendering [#352]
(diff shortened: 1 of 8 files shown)


sunng87:
We have tests for another issue that the return value of subexpression should be treated as literal. This will be fixed in next patch.

PR #364, opened 2020-07-20
Enable block parameters for derived values in `each` helper

### LINKED ISSUE #343: built-in helper `each` could not work with derived JSON value as parameters
Running the following code snippet

```rust
//# handlebars = "3.1.0"
//# serde_json = "*"

#[macro_use]
extern crate serde_json;

use handlebars::Handlebars;
use std::error::Error;

fn main() -> Result<(), Box<dyn Error>> {
    let data = json!(4);
    let hb = Handlebars::new();

    let template = "{{#each [0,1,2] as |i|}}{{i}}{{/each}}";
    println!("{}", hb.render_template(template, &data)?);
    Ok(())
}
```

Produces output `444` instead of `012`

chaserhkj:
Fixes #343

### DIFF
--- a/src/helpers/helper_each.rs
+++ b/src/helpers/helper_each.rs
@@ -444,10 +444,8 @@ mod test {
         assert_eq!("12", rendered);
     }
 
-
-
     #[test]
-    fn test_derived_value_as_block_params() {
+    fn test_derived_array_as_block_params() {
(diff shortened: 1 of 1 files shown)


sunng87:
Thanks @chaserhkj ! 

This patch enables access to derived value via block param. In #355 I'm going to add support for this completely by stored the derived value into `BlockContext`. I might change some of this impl in future.

PR #366, opened 2020-07-26
Add more escape chars

### LINKED ISSUE #360: More html espace chars
In addition to `<`, `>`, `&`, `"`, handlebars spec also mentioned more characters should be escaped.

sunng87:
Fixes #360

### DIFF
--- a/src/support.rs
+++ b/src/support.rs
@@ -39,6 +39,7 @@ pub mod str {
         }
     }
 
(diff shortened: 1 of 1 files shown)


sunng87:
@Keruspe in this patch `=` will be replaced by `"&#x3D;"`. I just ping you to ensure if it will affect your generator. If you are using handlebars to generate non-html content, I would recommend you to register a no_escape function to prevent such html escaping.

Keruspe:
@sunng87 I already register the no_escape fn, but thanks for the ping!

PR #383, opened 2020-09-26
Added `register_local_helper_rc` to `RenderContext`.

ExPixel:
Since `register_local_helper` and `get_local_helper` already make use of `Rc` in their signatures I figured there wouldn't be any harm in providing a way to create helpers using `Rc` directly without the reallocation required by `Rc::from_box`. It also allows using the same helper definition for multiple helpers with different names.

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -199,6 +199,18 @@ impl<'reg: 'rc, 'rc> RenderContext<'reg, 'rc> {
             .insert(name.to_string(), def.into())
     }
 
+    /// Register a reference counted helper in this render context.
+    ///
+    /// If there was a local helper with the same name present, it will be replaced
+    /// with the new local helper and returned, otherwise this will return `None`.
+    pub fn register_local_helper_rc(
+        &mut self,
+        name: &str,
+        def: Rc<dyn HelperDef + 'rc>,
+    ) -> Option<Rc<dyn HelperDef + 'rc>> {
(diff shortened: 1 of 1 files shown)


sunng87:
Thank you for your patch! Instead of creating a new method for this, I'm +1 for your idea of using generics. I'm working on 4.0 so even a little break change is acceptable at the moment.

ExPixel:
So I went ahead and just made it generic in a bit of a different way. This should still allow `Rc<dyn HelperDef>` (I had to implement that one manually) and `Box<dyn HelperDef>` but should also just allow passing closures or anything else that implements `HelperDef`.

sunng87:
This looks great to me! Please go ahead.

ExPixel:
Alright, so upon further inspection (and after more sleep) this doesn't really do what I set out for it to. The `Rc::from(T)` call just just going to produce an `Rc<Rc<T>>`. Without specialization there isn't really a way for me to make a set of traits that will allow `T where T: HelperDef`, `Box<T> where T: HelperDef` and `Rc<T> where T: HelperDef` without extra allocations for `Box<T>` in `Handlebars::register_helper` and for `Rc<T>` in `RenderContext::register_local_helper`. Requiring `Into<Rc<dyn HelperDef>>` works but it messes up type inference and makes passing `Box::new(new_helper)` require some inconvenient type gymnastics. Going to have to mess around with trait bounds a bit more to get something that works decently well.

PR #397, opened 2020-11-26
Update registry.rs

### LINKED ISSUE #396: Registry impl Clone?
Is it possible for the registry to implement clone so that we can clone it easily and pass it around?

rakshith-ravi:
Added Clone to registry.

Closes #396

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -51,6 +51,7 @@ pub fn no_escape(data: &str) -> String {
 /// The single entry point of your Handlebars templates
 ///
 /// It maintains compiled templates and registered helpers.
+#[derive(Clone)]
 pub struct Registry<'reg> {
     templates: HashMap<String, Template>,
     helpers: HashMap<String, Box<dyn HelperDef + Send + Sync + 'reg>>,


rakshith-ravi:
Okay yeah this doesn't seem to fix it. Let me try something else

sunng87:
Can you use a `Arc` for this?

rakshith-ravi:
Yup. That's what I'm currently using locally. But I'm trying to get this to Clone directly so that we can make multiple copies of the registry that can work independent of each other. I'm trying to use generic parameters. Will update the PR once I get something running

sunng87:
I'm afraid generic parameters may not work for this. Handlebars-rust uses a lot of dynamic dispatch in render and helper API, which makes `helpers` field not clone-able. I will think about this too.

sunng87:
By the way, I have made `Registry` clone-able in #395

rakshith-ravi:
Ahh, I see you made it into an `Rc`, to make it clone-able. While this works well, it is a little difficult if we need to have multiple copies of the `Registry` that can be cloned and modified independently. Sure, this can be a solution for now, don't get me wrong. I really appreciate the effort you're taking to fix this, and I would like to thank you for that. But if it's okay, I would like to take some more time to look for another way to make it really cloneable.

SuperCuber:
@rakshith-ravi correct me if I'm wrong, but I think right now the registry can be modified independently - it's just the HelperDefs themselves that cannot be. The map itself is cloned so you can just replace the helpers with other ones.

PR #403, opened 2020-12-25
Fix punctuation in errors built with `RenderError::from_error`.

mkantor:
These error messages ended up with a period + colon. For example:

```
Error on accessing array/vector with string index.: invalid digit found in string
```

```
Error with parsing template.: Template error: invalid handlebars syntax.
```


sunng87:
Thanks!

Issue #358, opened 2020-07-17
Remove decorator from public API


sunng87:
Handlebarsjs has deprecated decorator API from public and will only use for internal features (partial/inline).

kolektiv:
Found this as I'm curious about inline (as it's the only template inheritance-like mechanism I can see, and the example with partials here implies that). What would be the alternative if this goes away? There doesn't seem to be an alternative?

PR #408, opened 2021-01-03
Resolve helper params lazily and more

sunng87:
This patch made helper parameters/hash resolution lazily to avoid unnecessary overhead. 

It also removes public api of decorator  #358


sunng87:
This patch didn't bring performance benefit as expected. Close and won't merge.

PR #413, opened 2021-02-12
Light cloning of templates

dusdanig:
Currently we are serving templates which are re-loadable
asynchronously.  There are also assets which are shared by the
templates and are accessible through an async read lock.

For rendering we need to register specific helpers which will alter
the response for each request. The solution would be to clone all
templates for each request.  Then add the specific helpers for that
situation. This is not ideal, because cloning all templates is slow.

So by adding an Arc to Templates, they can be shared between all
handlers building responses. Cloning is somewhat lighter with this
change. It does break the get_templates call on registry.

The other solution I have thought of was to add a generic type to the
render call, which can be used by the helpers. This can then default
to () for common cases.

Third solution would be to separate the template management from the
registry of helpers. In that case the templates can be put into a
custom structure, and would aid custom reloading schemes. E.g. loading
of the templates and then add that to a second structure holding the
helpers, then render.  The registry is a composition of these two
structures.

Asynchronous helpers would also fix this for us. But in the mean time
this would help.

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -54,7 +54,7 @@ pub fn no_escape(data: &str) -> String {
 /// It maintains compiled templates and registered helpers.
 #[derive(Clone)]
 pub struct Registry<'reg> {
-    templates: HashMap<String, Template>,
+    templates: HashMap<String, Arc<Template>>,
 
     helpers: HashMap<String, Arc<dyn HelperDef + Send + Sync + 'reg>>,
     decorators: HashMap<String, Arc<dyn DecoratorDef + Send + Sync + 'reg>>,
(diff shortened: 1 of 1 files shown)


sunng87:
While this patch is certainly acceptable, but I'm thinking about other solution for your case, to avoid making every reference in rust an `Arc`.

In the upcoming 4.0 release, we will have a `dev_mode` that load template from disk every time when it's accessed. We have a structure called `TemplateSource` to track where the template was loaded. It's pretty much like your solution 3. Currently `TemplateSource` is an internal mechanism for `dev_mode` only. I suggest you to use it for your case and make it public where needed.

dusdanig:
Thanks for taking the time to answer.

sunng87:
@dusdanig I attempted to make `Source` a public API in #415  and add a new API to registry for source registration. If it works, you can just implement your own `Source`, perhaps with some cache capability.

PR #415, opened 2021-02-16
Publish source api

sunng87:


mkantor:
This API would be useful for me, so I brought this branch up-to-date with #444.

src/registry.rs
```diff
@@ -467,14 +492,23 @@ impl<'reg> Registry<'reg> {
         self.decorators.get(name).map(|v| v.as_ref())
     }
 
-    /// Return all templates registered
-    pub fn get_templates(&self) -> &HashMap<String, Template> {
-        &self.templates
+    /// Return all  registered template names
+    pub fn get_template_names(&self) -> Vec<String> {
```
sunng87, line 496: Sad that we didn't catch the 4.0 release train to ship this break change.

With the `Source` API, the original `get_templates() -> &HashMap<String, Template>` has lost its meaning

sunng87:
@mkantor may I know your use case for this feature?

mkantor:
> @mkantor may I know your use case for this feature?

My use case is [to set up a registry from already-open `File`s](https://github.com/mkantor/operator/blob/0.1.2/src/content/content_engine.rs#L238-L297).

With handlebars-rust 3.x I'm using [the old `register_template_source` method](https://github.com/sunng87/handlebars-rust/blob/v3.5.5/src/registry.rs#L278-L293) which accepts any [`Read`](https://doc.rust-lang.org/std/io/trait.Read.html). With the `Source` API published in this PR I could create a `impl Source for File` and use the new `register_template_source` from this changeset. _(...I'm now realizing this might not be precisely what I want, since I think I'd have to clone the `File` in `impl Source for File`. I guess my ideal API would be something closer to the old one where the source may be consumed/mutated. For my use case it's fine to give away ownership of the `File`.)_

If this doesn't get merged I'll probably use the path with `register_template_file` and have handlebars-rust open a new `File`, or maybe I'll just read the file to a `String` and use `register_template_string`, but I was kind of curious to try out dev mode.

mkantor:
I ended up using `register_template_file` instead, which seems to be working nicely (https://github.com/mkantor/operator/pull/58). I don't think I have a strong use case for the source API after all.

PR #418, opened 2021-02-18
(doc) mention no_escape() in the "Escaping" section

vadz:
This seems to be by far the most popular argument when calling
register_escape_fn(), judging from GitHub code search, and so it seems
worthwhile to mention it here to avoid the need for the people to write
their own trivial escape function doing nothing only to discover that
such function already exists in Handlebars itself.

---

A really trivial PR just to save some time to the others who might also waste time as I did before discovering `no_escape()`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -237,7 +237,7 @@
 //!
 //! #### Escaping
 //!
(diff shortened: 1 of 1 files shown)


sunng87:
Thanks!

Issue #427, opened 2021-05-01
Panic when generating error object for malformed template


5225225:
Reproduction code:

```rust
fn main() {
    let s = "{{#>(X)}}{{/X}}";
    let tpl = handlebars::Handlebars::new();
    let _ = tpl.render_template(&s, &Vec::<()>::new());
}
```

Looks like the actual value of what is in `X` there doesn't matter (and they don't need to be the same character, but can be).

The rest I couldn't remove without the panic not happening.

Stack trace (when testing the above on the latest version, commit 39589d9):

This happens on both the crates.io version as well as the latest version from git.

https://github.com/sunng87/handlebars-rust/blob/39589d9fb5f84c492504e7c3abeb7264f9fcd68a/src/template.rs#L660-L665 is where the panic happens, on line 662

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/jess/.cargo/git/checkouts/handlebars-rust-6b5b431bf9aaa55b/39589d9/src/template.rs:662:62
stack backtrace:
   0: rust_begin_unwind
             at /rustc/42816d61ead7e46d462df997958ccfd514f8c21c/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/42816d61ead7e46d462df997958ccfd514f8c21c/library/core/src/panicking.rs:92:14
   2: core::panicking::panic
             at /rustc/42816d61ead7e46d462df997958ccfd514f8c21c/library/core/src/panicking.rs:50:5
   3: core::option::Option<T>::unwrap
             at /home/jess/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:386:21
   4: handlebars::template::Template::compile
             at /home/jess/.cargo/git/checkouts/handlebars-rust-6b5b431bf9aaa55b/39589d9/src/template.rs:662:45
   5: handlebars::registry::Registry::render_template_to_write
             at /home/jess/.cargo/git/checkouts/handlebars-rust-6b5b431bf9aaa55b/39589d9/src/registry.rs:555:19
   6: handlebars::registry::Registry::render_template
             at /home/jess/.cargo/git/checkouts/handlebars-rust-6b5b431bf9aaa55b/39589d9/src/registry.rs:523:9
   7: scratchPMtZAxU7Y::main
             at ./src/main.rs:4:13
   8: core::ops::function::FnOnce::call_once
[...]

PR #430, opened 2021-05-03
Fix out of bounds index when parsing JSON path

5225225:
Fixes a bug I found when running the fuzzer for a bit after #427 was fixed

Not sure if the test here is in the right place or if you want me to put it in with the other test function.

Ran the fuzzer for a while after this and didn't see any issues. I'll make a new PR for the fuzzing target (or just lump it in here) if you want, it uses [`cargo-fuzz`](https://github.com/rust-fuzz/cargo-fuzz).


sunng87:
> Ran the fuzzer for a while after this and didn't see any issues. I'll make a new PR for the fuzzing target (or just lump it in here) if you want, it uses cargo-fuzz.

Thank you! Please make a PR for that

PR #431, opened 2021-05-03
Add fuzzing harness

5225225:
I get *really slow* test cases with this, so I'm guessing there's some non-linear time parsing that can cause extreme slowdowns with this.

You can specify a max length when running the fuzzer, I've found a length of 64 doesn't run into those issues and lets me test for panics, but it might be worth having a look at the slow cases, I've not been able to get a clean test case for the slowdown issue.

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,4 @@
+
+target
+corpus
+artifacts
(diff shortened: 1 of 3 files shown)


sunng87:
Thank you!

sunng87:
I will look into the slowdown issue. Maybe a profiler could help with this case.

PR #445, opened 2021-06-08
Fix: Calling each on empty iterator without else block

matthiasbeyer:
I added this test because updating from `3` to `4` broke my software in a weird way, where a variable was noted as not there, but it was there.

So here's a test to check whether things still work with empty iterables in the `#each` helper, because I was not able to get my local clone to run through `cargo test` (some weird issue with pprof not compiling... I don't know. The quick fix is to push this through github actions).

### DIFF
--- a/src/helpers/helper_each.rs
+++ b/src/helpers/helper_each.rs
@@ -80,7 +80,9 @@ impl HelperDef for EachHelper {
 
         match template {
             Some(t) => match *value.value() {
-                Json::Array(ref list) if !list.is_empty() => {
+                Json::Array(ref list)
+                    if !list.is_empty() || (list.is_empty() && h.inverse().is_none()) =>
+                {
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
So it seems that it still works here. Why the hell?

matthiasbeyer:
If this works, I will rebase to make the history clean (and rename the PR to something more appropriate).

matthiasbeyer:
Closes #446

sunng87:
Good catch. Thank you for the fix.

matthiasbeyer:
Yeah, nice!

May I kindly ask for a patch release? :heart: :pray:

sunng87:
4.0.1 uploaded

matthiasbeyer:
:partying_face: :tada: :heart:

PR #448, opened 2021-06-20
Empty line stripping fix

### LINKED ISSUE #447: Unexpected newline removal
After https://github.com/sunng87/handlebars-rust/pull/404, it seems lines are being removed a bit too aggressively.  If I have a template like this:

```
{{#if a}}x{{/if}}
y
```

I would expect it to render like this, if the data is `{"a": true}`:

```
x
y
```

This is what happens in handlebars.js, which I tested on [tryhandlebarsjs.com](http://tryhandlebarsjs.com/).

In handlebars-rust 4, this renders as:

```
xy
```

sunng87:
Fixes #447

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -403,26 +403,15 @@ impl Template {
     }
 
     // test if a statement is followed by a rawstring
-    fn is_adjust_to_rawstring<'a, I>(it: &mut Peekable<I>) -> bool
-    where
-        I: Iterator<Item = Pair<'a, Rule>>,
-    {
-        if let Some(pair) = it.peek() {
-            return pair.as_rule() == Rule::raw_text;
-        }
-
-        false
+    fn is_adjust_to_newline(source: &str) -> bool {
+        grammar::starts_with_empty_line(source)
     }
 
-    fn is_standalone_statement<'a, I>(
-        template_stack: &VecDeque<Template>,
-        it: &mut Peekable<I>,
-    ) -> bool
-    where
-        I: Iterator<Item = Pair<'a, Rule>>,
-    {
-        Template::check_previous_rawstring_for_newline(template_stack)
-            && Template::is_adjust_to_rawstring(it)
+    fn is_standalone_statement(template_stack: &VecDeque<Template>, source: &str) -> bool {
+        let r1 = Template::check_previous_rawstring_for_newline(template_stack);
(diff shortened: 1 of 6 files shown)


sunng87:
This cannot easily be fixed from pest parser level. I'm looking into template parser to resolve this.

Rules for whitespace auto elimination is to check if the directive `{{# xxx}}`` and ``{{/ xxx}}` is holding a whole line, with leading and trailing whitespaces counted, and then remove the trailing `NEWLINE` and keep the leading one.

sunng87:
The original JavaScript version: https://github.com/handlebars-lang/handlebars.js/pull/787

src/template.rs
```diff
@@ -389,10 +389,36 @@ impl Template {
         }
     }
 
+    fn check_previous_rawstring_for_newline(template_stack: &VecDeque<Template>) -> bool {
+        let t = template_stack.front().unwrap();
+        if let Some(el) = t.elements.get(0) {
```
sunng87, line 394: get last element of the template

sunng87 (commented):
leading whitespace has to be removed too

src/template.rs
```diff
@@ -442,22 +480,14 @@ impl Template {
...
-                !matches!(
-                    p.as_rule(),
-                    Rule::escape
-                        | Rule::brackets_open
-                        | Rule::brackets_close
-                        | Rule::double_brackets_open
-                        | Rule::double_brackets_close
-                        | Rule::comment_brackets_open
-                )
+                !matches!(p.as_rule(), Rule::escape | Rule::helper_block)
```
sunng87, line 490: remove `Rule::helper_block`

src/template.rs
```diff
@@ -433,6 +469,8 @@ impl Template {
         let mut template_stack: VecDeque<Template> = VecDeque::new();
 
         let mut omit_pro_ws = false;
+        // flag for tracking standalone statement like {{#if}} and {{/if}}
```
sunng87, line 472: update comment to clarify its purpose

src/template.rs
```diff
@@ -554,6 +598,13 @@ impl Template {
                         }
                         omit_pro_ws = exp.omit_pro_ws;
 
+                        // standalone line check part 1, for the leading whitespaces and newline
```
sunng87, line 601: update comment to its latest  behaviour

Issue #449, opened 2021-06-25
Strict mode and boolean helpers


twittner:
Would it make sense to error if parameter values of boolean helpers are missing in strict mode? For example I would like this test to pass:

```diff
diff --git a/tests/subexpression.rs b/tests/subexpression.rs
index d1803b1..1086508 100644
--- a/tests/subexpression.rs
+++ b/tests/subexpression.rs
@@ -7,6 +7,17 @@ use std::sync::Arc;

 use handlebars::{Context, Handlebars, Helper, HelperDef, RenderContext, RenderError, ScopedJson};

+#[test]
+fn test_strict_mode() {
+    let mut hbs = Handlebars::new();
+    hbs.set_strict_mode(true);
+
+    let data = json!({"a": 1});
+
+    assert!(hbs.render_template("{{#if (eq a 1)}}Success{{else}}Failed{{/if}}", &data).is_ok());
+    assert!(hbs.render_template("{{#if (eq z 1)}}Success{{else}}Failed{{/if}}", &data).is_err())
+}
+
 #[test]
 fn test_subexpression() {
     let hbs = Handlebars::new();

```

where `z` is unbound. Would something like the following be a sensible solution? It may break existing templates.

```diff
diff --git a/src/macros.rs b/src/macros.rs
index fde09d9..14cb015 100644
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -51,7 +51,7 @@ macro_rules! handlebars_helper {
             fn call_inner<'reg: 'rc, 'rc>(
                 &self,
                 h: &$crate::Helper<'reg, 'rc>,
-                _: &'reg $crate::Handlebars<'reg>,
+                r: &'reg $crate::Handlebars<'reg>,
                 _: &'rc $crate::Context,
                 _: &mut $crate::RenderContext<'reg, 'rc>,
             ) -> Result<$crate::ScopedJson<'reg, 'rc>, $crate::RenderError> {
@@ -59,7 +59,13 @@ macro_rules! handlebars_helper {

                 $(
                     let $name = h.param(param_idx)
-                        .map(|x| x.value())
+                        .and_then(|x| {
+                            if r.strict_mode() && x.is_value_missing() {
+                                None
+                            } else {
+                                Some(x.value())
+                            }
+                        })
                         .ok_or_else(|| $crate::RenderError::new(&format!(
                             "`{}` helper: Couldn't read parameter {}",
                             stringify!($struct_name), stringify!($name),
```

[...]

PR #451, opened 2021-07-01
Error if boolean helper params are missing.

twittner:
Unbound parameters of test expressions now cause an error if strict mode is enabled. For context see issue #449.


sunng87:
Thanks!

PR #454, opened 2021-07-12
Export `push_partial_block` and `pop_partial_block`

mkantor:
This allows custom helpers to manipulate the `@partial-block` variable.

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -177,11 +177,11 @@ impl<'reg: 'rc, 'rc> RenderContext<'reg, 'rc> {
         self.inner_mut().partials.insert(name, partial);
     }
 
-    pub(crate) fn push_partial_block(&mut self, partial: &'reg Template) {
+    pub fn push_partial_block(&mut self, partial: &'reg Template) {
         self.inner_mut().partial_block_stack.push_front(partial);
     }
 
-    pub(crate) fn pop_partial_block(&mut self) {
+    pub fn pop_partial_block(&mut self) {
         self.inner_mut().partial_block_stack.pop_front();
     }
 


sunng87:
May I know your use case for this? Since `@partial-block` is an internal mechanism in handlebars. Exposing these API  may allow developers to change its meaning of it

mkantor:
> Exposing these API may allow developers to change its meaning of it

That's my goal. :smile:

I have a custom helper that can render other handlebars templates (with additional functionality), and want to set `@partial-block` so it behaves more like `>` (https://github.com/mkantor/operator/issues/10#issuecomment-880379384).

Here's the commit where I'd like to use `push_partial_block`: https://github.com/mkantor/operator/commit/bd80936ad85a3df3a4670702341879bea2bfa307#diff-6d2087fee5a1b9f51fa4201717f8bbc6abea61b7a185437a0bb50da97bb00c0e

sunng87:
@mkantor Thanks for your clarification. How about using `set_partial` from `RenderContext` to register the template as a partial in your helper? The only drawback I think is you cannot use `@partial-block` this way but it makes sense because `@partial-block` has it own meaning within handlebars system.

mkantor:
@sunng87 Sorry for the slow response!

> How about using `set_partial` from `RenderContext` to register the template as a partial in your helper? The only drawback I think is you cannot use `@partial-block` this way but it makes sense because `@partial-block` has it own meaning within handlebars system.

I could use `set_partial` with a different name (that's what I did while prototyping), but I was hoping to specifically use `@partial-block` to make it easier to migrate between `>` and my `get` helper. My goal is to not require any changes in the included template when you do that.

To me this doesn't feel much different from a helper that sets variables like `@index` or `@key`, but I may be missing something.

For what it's worth [handlebars.js allows this](https://handlebarsjs.com/playground.html#format=1&currentExample=%7B%22template%22%3A%22%7B%7B%23test%7D%7D%7B%7B%3E%40partial-block%7D%7D%7B%7B%2Ftest%7D%7D%22%2C%22partials%22%3A%5B%5D%2C%22input%22%3A%22%7B%7D%22%2C%22output%22%3A%22hello%22%2C%22preparationScript%22%3A%22Handlebars.registerHelper('test'%2C%20options%20%3D%3E%20%7B%5Cn%20%20Handlebars.registerPartial('%40partial-block'%2C%20'hello')%5Cn%20%20const%20output%20%3D%20options.fn(%7B%7D)%5Cn%20%20Handlebars.unregisterPartial('%40partial-block')%5Cn%20%20return%20output%5Cn%7D)%22%2C%22handlebarsVersion%22%3A%224.7.7%22%7D) (there it's much worse because the global `Handlebars` object gets mutated).

PR #465, opened 2021-08-29
fix link to django doc for template inheritance

peddermaster2:
The old django 1.9 doc is no longer available.  
https://docs.djangoproject.com/en/1.9/ref/templates/language/#template-inheritance


sunng87:
Thank you!

PR #479, opened 2021-12-05
Fix handling of identifiers which start with numbers

### LINKED ISSUE #450: Helper arguments starting with a number are split in two parts
When a helper parameter starts with a number, but contain something else after, it will be split at the first non-digit character.

For example `{{helper 100_000}}` will result in an agument sequence of `[PathAndJson { relative_path: None, value: Constant(Number(100)) }, PathAndJson { relative_path: Some("_000"), value: Missing }]`

I would expect one of:
* String `"100_000"`
* Number `100000`
* Syntax error

But truncating here seems like a bug.

Full example:

```rust
use std::collections::BTreeMap;
use handlebars::{Handlebars, RenderContext, Helper, Context, JsonRender, HelperResult, Output};

fn echo_helper (h: &Helper, _: &Handlebars, _: &Context, rc: &mut RenderContext, out: &mut dyn Output) -> HelperResult {
    println!("params={:?}", h.params());
    let param = h.param(0).unwrap();
    out.write(param.value().render().as_ref())?;
    Ok(())
}

fn main() {
  // create the handlebars registry
  let mut handlebars = Handlebars::new();
  handlebars.register_helper("echo", Box::new(echo_helper));

  // register the template. The template string will be verified and compiled.
  let source = "amount: {{echo 100_000}}";
  assert!(handlebars.register_template_string("t1", source).is_ok());

  // Prepare some data.
  //
  // The data type should implements `serde::Serialize`
  let data: BTreeMap<String, String> = BTreeMap::new();
  assert_eq!(handlebars.render("t1", &data).unwrap(), "amount: 100_000");
}
```

Output with handlebars-rust 4.0.1:
```
params=[PathAndJson { relative_path: None, value: Constant(Number(100)) }, PathAndJson { relative_path: Some("_000"), value: Missing }]
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `"amount: 100"`,
 right: `"amount: 100_000"`', src/main.rs:24:3
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

mkantor:
Fixes #450.

### DIFF
--- a/src/grammar.pest
+++ b/src/grammar.pest
@@ -14,7 +14,7 @@ literal = { string_literal |
 
 null_literal = @{ "null" ~ !symbol_char }
 boolean_literal = @{ ("true"|"false") ~ !symbol_char }
-number_literal = @{ "-"? ~ ASCII_DIGIT+ ~ "."? ~ ASCII_DIGIT* ~ ("E" ~ "-"? ~ ASCII_DIGIT+)? }
+number_literal = @{ "-"? ~ ASCII_DIGIT+ ~ "."? ~ ASCII_DIGIT* ~ ("E" ~ "-"? ~ ASCII_DIGIT+)? ~ !symbol_char }
 json_char_double_quote = {
     !("\"" | "\\") ~ ANY
     | "\\" ~ ("\"" | "\\" | "/" | "b" | "f" | "n" | "r" | "t")
--- a/src/render.rs
+++ b/src/render.rs
@@ -1117,3 +1117,37 @@ fn test_zero_args_heler() {
         "Output name: first_name not resolved"
     );
 }
+
+#[test]
+fn test_identifiers_starting_with_numbers() {
+    let mut r = Registry::new();
+
+    assert!(r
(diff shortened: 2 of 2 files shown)


mkantor:
[Here's an example of this working in handlebars.js](https://handlebarsjs.com/playground.html#format=1&currentExample=%7B%22template%22%3A%22%7B%7B%23if%200a%7D%7Dtrue%7B%7B%2Fif%7D%7D%22%2C%22partials%22%3A%5B%5D%2C%22input%22%3A%22%7B%5C%220a%5C%22%3A%20true%7D%22%2C%22output%22%3A%22true%22%2C%22preparationScript%22%3A%22%22%2C%22handlebarsVersion%22%3A%224.7.7%22%7D). Without this change handlebars-rust behaves differently (emitting no output in that example).

sunng87:
Thank you @mkantor for the patch! 

Also feel free to share your ideas with this library, I'm looking for co-maintainers of this project.

PR #484, opened 2021-12-28
Rust embedded support

### LINKED ISSUE #478: rust-embed integration
https://github.com/pyros2097/rust-embed 

Rust-embed embeds external resources into rust binary. It would be nice to provide first-class citizen support for the library so that we can `register_embedded_template` like we did with directory.

sunng87:
Fixes #478

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -30,6 +30,9 @@ use rhai::Engine;
 #[cfg(feature = "script_helper")]
 use crate::helpers::scripting::ScriptHelper;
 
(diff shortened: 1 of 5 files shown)


src/registry.rs
```diff
@@ -322,6 +325,25 @@ impl<'reg> Registry<'reg> {
         Ok(())
     }
 
+    /// Register templates using a RustEmbed type
```
sunng87, line 328: Improve docs for `register_embed_templates`

PR #500, opened 2022-04-06
Refactor dir_source to be more ideomatic

Fallen4Eyes:
This pull request was originally for allowing extensions to not be prefixed by a dot which lead to a general refactor of dir_source.

The main change is that tpl_extension in register_templates_directory doesn't need a dot in the front and non-utf8 file names will not be registered unlike before where mangled strings will be registered.

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -306,22 +306,41 @@ impl<'reg> Registry<'reg> {
         };
 
         let walker = WalkDir::new(dir_path);
-        let dir_iter = walker.min_depth(1).into_iter().filter_map(|e| {
-            let e = e.ok()?;
-            let path = e.path();
-            path.extension()
-                .and_then(|extension| {
-                    if extension == tpl_extension {
-                        path.file_stem()
-                    } else {
-                        None
-                    }
-                })
-                .and_then(|stem| stem.to_str())
-                .map(|stem| (path.to_owned(), stem.to_owned()))
-        });
+        let dir_iter = walker
+            .min_depth(1)
+            .into_iter()
+            .filter_map(|e| e.ok().map(|e| e.into_path()))
+            // Checks if extension matches
+            .filter(|tpl_path| {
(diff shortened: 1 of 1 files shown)


sunng87:
@Fallen4Eyes Thank you for the patch! There is a test case failure, please check.

src/registry.rs
```diff
@@ -307,39 +290,38 @@ impl<'reg> Registry<'reg> {
...
+                        None
+                    }
+                })
+                .and_then(|stem| stem.to_str())
+                .map(|stem| (path.to_owned(), stem.to_owned()))
+        });
 
-            let tpl_name = &tpl_file_path[prefix_len..tpl_file_path.len() - tpl_extension.len()];
-            // replace platform path separator with our internal one
-            let tpl_canonical_name = tpl_name.replace(path::MAIN_SEPARATOR, "/");
```
sunng87, line 342: this part of logic may be lost during the refactor

src/registry.rs
```diff
@@ -94,21 +92,6 @@ impl<'reg> Default for Registry<'reg> {
...
-fn filter_file(entry: &DirEntry, suffix: &str) -> bool {
-    let path = entry.path();
-
-    // ignore hidden files, emacs buffers and files with wrong suffix
-    !path.is_file()
-        || path
-            .file_name()
-            .map(|s| {
-                let ds = s.to_string_lossy();
-                ds.starts_with('.') || ds.starts_with('#') || !ds.ends_with(suffix)
```
sunng87, line 107: file names starts with `.` and `#` should be skipped.

Fallen4Eyes:
I made changes in order to preserve the functionality of the old code as I misunderstood it's old functionality. (making sure I enabled the correct features when running the tests helped too)

sunng87 (approved):

sunng87:
@Fallen4Eyes Thank you!

PR #508, opened 2022-05-26
Make registry public

kareemk:
This is needed to be able to assign the registry to a static ref using lazy_static as the type cannot be inferred and must be declared.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -412,7 +412,7 @@ mod json;
 mod local_vars;
 mod output;
 mod partial;
-mod registry;
+pub mod registry;
 mod render;
 mod sources;
 mod support;


sunng87:
hi @kareemk 

I was able to use lazy_static like:

```rust
use handlebars::Handlebars;

use lazy_static::lazy_static;

lazy_static! {
    static ref REG: Handlebars<'static> = Handlebars::new();
}

```

Could you provide me an example for the case that you try to resolve?

kareemk:
Ok, that worked. I'm still new to Rust and from the type hints on `Handlebars::new()` showing the returned type to be a `handlebars::registry::Registry` that I needed to use that type. Thank you.

PR #522, opened 2022-07-30
feat: Add in trait Output in write_fmt to get access to write! macro

darnuria:
Upside: It may help (or not) to avoid allocation, rustc, is not able every time able to discard format allocations.
Downsite API change, may not be worth it.

Edit: 
- had some personal mixed results on `cargo bench` so I removed the comments because it was mostly noise to retain the real addition of this PR: adding `write_fmt` to `Output` trait or not, totally ok if it's closed as "not worth the maintaining effort" or "not in the scope of the project" :)

Have a great day

### DIFF
--- a/examples/decorator.rs
+++ b/examples/decorator.rs
@@ -23,8 +23,7 @@ fn format_helper(
     let param = h
         .param(0)
         .ok_or(RenderError::new("Param 0 is required for format helper."))?;
-    let rendered = format!("{} pts", param.value().render());
-    out.write(rendered.as_ref())?;
+    write!(out, "{} pts", param.value().render())?;
     Ok(())
 }
 
@@ -51,8 +50,7 @@ fn format_decorator(
                 let param = h
                     .param(0)
                     .ok_or(RenderError::new("Param 0 is required for format helper."))?;
-                let rendered = format!("{} {}", param.value().render(), suffix);
-                out.write(rendered.as_ref())?;
+                write!(out, "{} {}", param.value().render(), suffix)?;
                 Ok(())
             },
         ),
--- a/src/helpers/mod.rs
+++ b/src/helpers/mod.rs
@@ -209,12 +209,9 @@ mod test {
             let v = h.param(0).unwrap();
 
             if !h.is_block() {
-                let output = format!("{}:{}", h.name(), v.value().render());
-                out.write(output.as_ref())?;
+                write!(out, "{}:{}", h.name(), v.value().render())?;
             } else {
-                let output = format!("{}:{}", h.name(), v.value().render());
-                out.write(output.as_ref())?;
-                out.write("->")?;
+                write!(out, "{}:{}->", h.name(), v.value().render())?;
                 h.template().unwrap().render(r, ctx, rc, out)?;
             };
             Ok(())
@@ -258,8 +255,7 @@ mod test {
                  _: &mut RenderContext<'_, '_>,
                  out: &mut dyn Output|
                  -> Result<(), RenderError> {
-                    let output = format!("{}{}", h.name(), h.param(0).unwrap().value());
-                    out.write(output.as_ref())?;
+                    write!(out, "{}{}", h.name(), h.param(0).unwrap().value())?;
                     Ok(())
                 },
             ),
@@ -273,8 +269,7 @@ mod test {
                  _: &mut RenderContext<'_, '_>,
                  out: &mut dyn Output|
                  -> Result<(), RenderError> {
-                    let output = format!("{}", h.hash_get("value").unwrap().value().render());
-                    out.write(output.as_ref())?;
+                    write!(out, "{}", h.hash_get("value").unwrap().value().render())?;
                     Ok(())
                 },
             ),
--- a/src/helpers/mod.rs
+++ b/src/helpers/mod.rs
@@ -208,12 +208,11 @@ mod test {
         ) -> Result<(), RenderError> {
             let v = h.param(0).unwrap();
 
-            if !h.is_block() {
-                write!(out, "{}:{}", h.name(), v.value().render())?;
-            } else {
-                write!(out, "{}:{}->", h.name(), v.value().render())?;
+            write!(out, "{}:{}", h.name(), v.value().render())?;
+            if h.is_block() {
+                out.write("->")?;
                 h.template().unwrap().render(r, ctx, rc, out)?;
-            };
+            }
             Ok(())
         }
     }
--- a/src/output.rs
+++ b/src/output.rs
@@ -11,7 +11,13 @@ pub trait Output {
     /// for backward compatibility and to avoid breakage the default implementation
     /// uses `format!` this may be not what you want.
     fn write_fmt(&mut self, args: std::fmt::Arguments<'_>) -> Result<(), IOError> {
-        self.write(&std::fmt::format(args))
+        // Check if there is nothing to format to avoid allocation on case like
+        // write!(out, "hey")?;
+        if let Some(content) = args.as_str() {
+            self.write(content)
+        } else {
+            self.write(&std::fmt::format(args))
+        }
     }
 }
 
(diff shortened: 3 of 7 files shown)


darnuria:
Force push due to `cargo fmt`.

darnuria:
Ok the benchmark are not really stable and are really sensitive to rerun.

retried after an other big calculation non related to handlebar and got 34% degradation between two run on master, running benchmark several times is misleading branch predictor and cache may not be discarded enough by the warm up.

darnuria:
Ok after cleaning the previous bench, getting a totally unrelated workload runned master then redo something else and redo bench on this branch got that:

master:
```
parse_template          time:   [22.264 µs 22.398 µs 22.557 µs]                            
Found 14 outliers among 100 measurements (14.00%)
  8 (8.00%) high mild
  6 (6.00%) high severe

render_template         time:   [4.2800 µs 4.2980 µs 4.3191 µs]                             
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe

large_loop_helper       time:   [355.85 µs 362.22 µs 369.56 µs]                              
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe

large_loop_helper_with_context_creation                                                                            
                        time:   [883.69 µs 891.26 µs 899.41 µs]
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) high mild
  5 (5.00%) high severe

large_nested_loop       time:   [401.87 µs 415.90 µs 432.87 µs]                              
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) high mild
  3 (3.00%) high severe
```
this branch
```
parse_template          time:   [22.694 µs 22.841 µs 23.031 µs]                            
Found 11 outliers among 100 measurements (11.00%)
  2 (2.00%) low mild
  4 (4.00%) high mild
  5 (5.00%) high severe

render_template         time:   [4.4405 µs 4.4720 µs 4.5071 µs]                             
Found 9 outliers among 100 measurements (9.00%)
  6 (6.00%) high mild
  3 (3.00%) high severe

large_loop_helper       time:   [362.27 µs 364.18 µs 366.49 µs]                              
Found 12 outliers among 100 measurements (12.00%)
  2 (2.00%) high mild
  10 (10.00%) high severe

large_loop_helper_with_context_creation                                                                            
                        time:   [823.18 µs 826.82 µs 830.95 µs]
Found 12 outliers among 100 measurements (12.00%)
  1 (1.00%) low mild
  8 (8.00%) high mild
  3 (3.00%) high severe

large_nested_loop       time:   [374.46 µs 375.85 µs 377.39 µs]                              
Found 10 outliers among 100 measurements (10.00%)
  8 (8.00%) high mild
  2 (2.00%) high severe
```

sunng87:
@darnuria I saw you created another Pull Request #523. Is this one still a WIP?

sunng87 (approved):

darnuria:
Openned it as draft because it change a public `Output` API if i am right.

sunng87:
It makes sense to me. Although our internal implementation doesn't use fmt but it could be a plus for custom helper developer.

To retain compatibility I suggest you to provide a default implementation of `write_fmt` using `format!`so it won't break current users code

src/output.rs
```diff
@@ -16,6 +17,10 @@ impl<W: Write> Output for WriteOutput<W> {
     fn write(&mut self, seg: &str) -> Result<(), IOError> {
         self.write.write_all(seg.as_bytes())
     }
+
+    fn write_fmt(&mut self, args: std::fmt::Arguments<'_>) -> Result<(), IOError> {
```
sunng87, line 21:
```suggestion
    fn write_fmt(&mut self, args: std::fmt::Arguments<'_>) -> Result<(), IOError> {
        if let Some(content) = args.as_str() {
            self.write(content)
        } else {
            self.write(&args.to_string())
        }
    }
```

Add a default implementation for compatibility

examples/decorator.rs
```diff
@@ -23,8 +23,7 @@ fn format_helper(
     let param = h
         .param(0)
         .ok_or(RenderError::new("Param 0 is required for format helper."))?;
-    let rendered = format!("{} pts", param.value().render());
-    out.write(rendered.as_ref())?;
```
sunng87, line 27: update to `write_fmt`

sunng87 (changes requested):

examples/decorator.rs
```diff
@@ -23,8 +23,7 @@ fn format_helper(
...
+    write!(out, "{} pts", param.value().render())?;
```
darnuria, line 26: Is that done here?

src/helpers/mod.rs
```diff
@@ -209,12 +209,9 @@ mod test {
             let v = h.param(0).unwrap();
 
             if !h.is_block() {
-                let output = format!("{}:{}", h.name(), v.value().render());
-                out.write(output.as_ref())?;
+                write!(out, "{}:{}", h.name(), v.value().render())?;
             } else {
-                let output = format!("{}:{}", h.name(), v.value().render());
```
darnuria, line 215: This one may be even smarter and move out of the if `write!(out, "{}:{}", h.name(), v.value().render())?` and just have the iff for 
```
out.write("->")?;
h.template().unwrap().render(r, ctx, rc, out)?;
```

src/output.rs
darnuria: If what I did is not better than your suggestion tel me and I will apply your's had just a quick idea who don't surprise users, but it may be nice to trigger a warning if used as a default impl (not familiar enough with attributes to suggest one)

Thanks for your review time, and sorry for my english I am not a native english writter.

src/helpers/mod.rs
darnuria: (it's a detail and must not block IHMO it don't hurt to stay as it is.)

src/output.rs
darnuria: (whops commited directly on the `trait` it's late sorry.

src/output.rs
```diff
@@ -6,6 +6,13 @@ use std::string::FromUtf8Error;
 /// Handlebars uses this trait to define rendered output.
 pub trait Output {
     fn write(&mut self, seg: &str) -> Result<(), IOError>;
+
+    /// Designed to be used with `write!` macro.
+    /// for backward compatibility and to avoid breakage the default implementation
+    /// uses `format!` this may be not what you want.
+    fn write_fmt(&mut self, args: std::fmt::Arguments<'_>) -> Result<(), IOError> {
+        self.write(&std::fmt::format(args))
+    }
```
darnuria, line 15: The default `impl` shall be here correct me if I'am wrong.

src/helpers/mod.rs
sunng87: Right. This can be simplified.

src/output.rs
sunng87: By calling `args.as_str()`, it could avoid unnecessary allocation if the content of format arguments has nothing to format.

examples/decorator.rs
sunng87: Sorry I just realized `write!` uses `write_fmt` under the hood.

src/output.rs
sunng87: > Sure, had a long day will I pushed may be better (not sure) added a little bit of documentation to explain why it's better to not stick to the default impl (maybe there is an attribute to warn people using the default impl.

Yes the default implementation is just to keep the code compile if our user has custom `Output`. It should be no bad than their current implementations. For any new `Output`, developer should seek for an optimized implementation.

darnuria: The line:
`            self.write(&args.to_string())` will just print out the format string. This will do no substitution? Tell me if i'am wrong, first time I dig into `std::fmt::Arguments`

sunng87:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fbc78b7be03b74df9d90d04eaaad33f4

It generates formatted string.

sunng87 (approved):

sunng87:
Looks good to me now. Thank you so much for your work @darnuria

darnuria:
@sunng87 thanks for the review time and for maintaining the crate, learned some bits about write/format internal that was interesting.

PR #525, opened 2022-08-12
Include `axum-template` on related projects

Altair-Bueno:
`axum-template` provides `handlebars` integration with Axum


sunng87:
Nice! Thank you!

PR #532, opened 2022-09-23
issue 529: simplify lifetime requirements

### LINKED ISSUE #529: Rendering a template requires it to live for as long as the registry
Currently the types in this crate allow Template objects to live outside of the registry, but the lifetime annotations of the render method make it very hard to actually implement a system like that. 

Is there a reason why a template that lives shorter than the registry where partials are stored couldn't be rendered?

I would need that for [sqlpage](https://github.com/lovasoa/sqlpage)

lovasoa:
Only type annotations needed to be fixed, all the code actually already supported rendering shorter-lived templates.  

This fixes #529 

This is a breaking change, but it makes this API useful in more contexts, and even simplifies it a bit with fewer lifetime annotations, so I think it's worth it.

@sunng87

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -68,9 +68,9 @@ pub struct BlockContext<'rc> {
     local_variables: LocalVars,
 }
 
-impl<'reg> BlockContext<'reg> {
+impl<'rc> BlockContext<'rc> {
     /// create a new `BlockContext` with default data
-    pub fn new() -> BlockContext<'reg> {
+    pub fn new() -> BlockContext<'rc> {
         BlockContext::default()
     }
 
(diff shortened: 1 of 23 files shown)


sunng87:
@lovasoa Thank you for your effect on experimenting on this issue. There are some lint/test issues to address. The change is welcomed if it works for our test cases.

lovasoa:
Thank you ! Sorry, I had forgotten to update the code that was behind feature flags, it's done now :)

sunng87:
It all passed. Thank you for this patch. It has reduced overall complexity of this library. I will prepare a 5.0 release with this, including other break changes.

lovasoa:
Thanks to you, @sunng87, for maintaining this library :smiley:

PR #533, opened 2022-09-24
Allow public access to local variables

lovasoa:
This makes LocalVars public and gives mutable access to it from BlockContext. I need this for [SQLPage](https://github.com/lovasoa/SQLpage)


sunng87 (approved):
LGTM

PR #545, opened 2022-11-10
make all modules public.

Bjohnson131:
It makes things easier on me as a developer. For example, I want to have a function that sets up a common handlebars instance that's used by a lot of functions, so I make this:

```rs
pub fn render_handlebars() -> handlebars::registry::Registry<'static> {
    let mut hbs = Handlebars::new();
    // [a bunch of configuring]
    return hbs;
}
```

but it's not currently possible to make this, since handlebars::registry is private, so I must copy and paste my code everywhere I could normally use a function.

Signed-off-by: Brice Johnson <brice.johnson@wheelsup.com>


sunng87:
Thank you but Handlebars is exact Registry re-exported with a different name.

PR #546, opened 2022-11-19
Able to get the missing variable path from error 

linw1995:
This allows developers to get the missing variable path from the error and produce a customized hint message to the frontend users.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,5 +1,5 @@
 use std::error::Error as StdError;
-use std::fmt::{self, Write};
+use std::fmt::{self, Display, Write};
 use std::io::Error as IOError;
 use std::num::ParseIntError;
 use std::string::FromUtf8Error;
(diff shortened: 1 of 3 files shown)


src/error.rs
```diff
@@ -72,6 +72,22 @@ impl From<TemplateError> for RenderError {
     }
 }
 
+#[derive(Debug)]
+pub struct MissingVariableError(pub String);
```
sunng87, line 76: You can use `derive(Error)` to simplify this, see https://docs.rs/thiserror/latest/thiserror/

sunng87 (commented):
also please check format and lint errors. Thanks!

src/error.rs
```diff
@@ -102,11 +119,10 @@ impl RenderError {
...
 
     pub fn strict_error(path: Option<&String>) -> RenderError {
-        let msg = match path {
-            Some(path) => format!("Variable {:?} not found in strict mode.", path),
-            None => "Value is missing in strict mode".to_owned(),
-        };
-        RenderError::new(msg)
+        match path {
+            Some(path) => RenderErrorReason::MissingVariable(path.to_owned()).into(),
+            None => RenderError::new("Value is missing in strict mode"),
```
sunng87, line 124: Because both cases are `MissingVariable`. How about change this block to 

```rust
RenderErrorReason::MissingVariable(path.map(|p| p.to_owned())).into()
```

And make `MissingVariable(String)` hold `MissingVariable(Option<String>)`

Issue #575, opened 2023-05-02
Can't pass single-quoted strings as parameters


nickvollmar:
# Reproduction
In `tests/helper_macro.rs`, within the function `test_macro_helper`, you can add the following assertion.
```rust
    assert_eq!(
        hbs.render_template("{{tag 'html'}}", &()).unwrap(),
        "&lt;html&gt;"
    );
```

```
thread 'test_macro_helper' panicked at 'called `Result::unwrap()` on an `Err` value: RenderError { desc: "`tag` helper: Couldn't convert parameter t to type `str`. It's Null as JSON. Got these params: [PathAndJson { relative_path: Some(\"'html'\"), value: Missing }]", template_name: None, line_no: Some(1), column_no: Some(1), cause: None, unimplemented: false }', tests/helper_macro.rs:77:52
```

# Cause
I believe the origin of this behavior is https://github.com/sunng87/handlebars-rust/blob/27a9a9a1ee015056a3f803b28f4b1d1e68f048de/src/template.rs#L292-L296

The single-quoted literal matches the `Rule::literal` parser rule, but it is not valid JSON, so it becomes a Name object.

As evidence of this, here is the data structure I get by parsing the template `{{tag 'html'}}`:
```
Template {
  name: None,
  elements: [
    Expression(HelperTemplate {
      name: Name("tag"),
      params: [Name("'html'")],
      hash: {},
      block_param: None,
      template: None,
      inverse: None,
      block: false
    })
  ],
  mapping: [TemplateMapping(1, 1)]
}
```

# Proposal
I think the way forward is to somehow extract more information from the pest parse results. 

Here is a rough sketch of how I think that logic might look. As you can see, I'm out of my depth here, but this passes all the tests!
```rust
Rule::literal => {
    let param = param.into_inner().next().unwrap();
    match param.as_rule() {
        Rule::string_literal => {
            let string_inner = param.into_inner().next().unwrap();
            // don't double-escape characters - see test_string_no_escape_422
            let double_quoted = format!("\"{}\"", string_inner.as_str());
            Parameter::Literal(Json::from_str(&double_quoted).unwrap())
        }
        _ => {
            Parameter::Literal(Json::from_str(param_span.as_str()).unwrap())
        }
    }
}
```

(In the suggestion above I removed the Name code path, since I don't think any other JSON data type would make sense as a Name.)

Thank you very much, for reading this report and for all your work on this project.

PR #577, opened 2023-05-05
Fix parameter passing for single-quoted string literals

nickvollmar:
Addresses issue https://github.com/sunng87/handlebars-rust/issues/575.

I added some more checks to the `test_string_no_escape_422` function, because it seemed to be testing closely related behavior. But if you want I could put them in a new `575` function instead.

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -288,11 +288,24 @@ impl Template {
                 Parameter::Path(Path::new(param_span.as_str(), path_segs))
             }
             Rule::literal => {
-                let s = param_span.as_str();
-                if let Ok(json) = Json::from_str(s) {
-                    Parameter::Literal(json)
-                } else {
-                    Parameter::Name(s.to_owned())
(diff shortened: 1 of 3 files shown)


src/template.rs
```diff
@@ -288,11 +288,26 @@ impl Template {
...
+                        // and then parse it as a JSON literal.
+                        let string_inner_single_quote = it.next().unwrap();
+                        let double_quoted = format!(
+                            "\"{}\"",
+                            string_inner_single_quote
+                                .as_str()
+                                .replace("\\'", "'")
+                                .replace('"', "\\\"")
+                        );
+                        Parameter::Literal(Json::from_str(&double_quoted).unwrap())
```
sunng87, line 308: is it 100% safe to do `unwrap` here? Otherwise we better raise an TemplateError here

sunng87:
@nickvollmar Thank you!

PR #578, opened 2023-05-10
fix: use std::result::Result instead of Result in handlebar_helper!() macro

meuter:
It is often customary to define a crate wide `Error` and `Result` type. Something like:

```
type Error = ...
type Result<T> = std::result::Result<T, Error>
```

If such `Result<T>` type is in scope when calling the `handlebar_helper!()` macro, compilation fails as the number of template argument does not match.

This commit modified macro to rely explicitely on `std::result::Result<>` instead of the whatever Result type is in scope when the macro is invoked.


sunng87:
Thank you!

Issue #587, opened 2023-06-13
Rendering arrays leave a trailing comma


aromeronavia:
### Problem
When rendering an array, it renders always a trailing comma, which I'm assuming is not the intended functionality, but instead it should be rendering a comma just for all the elements prior to the last
https://github.com/sunng87/handlebars-rust/blob/23ca8d76bee783bf72f627b4c4995d1d11008d17/src/json/value.rs#L110-L130

## Example
We want to render the following array: [1, 2, 3]

## Expected Result
It gets rendered as `[1, 2, 3]`

## Actual Result
It gets rendered as `[1, 2, 3, ]`

PR #588, opened 2023-06-13
Render arrays without trailing commas

aromeronavia:
## Problem
Arrays are being rendered with trailing commas
Related to #587 

## Solution
Verify when rendering an array that last element being rendered does not include a comma and space

## Added tests ✅


sunng87:
Nice! Thank you very much!

PR #594, opened 2023-07-25
Added TemplateErrorReason to exported types

Julian-Alberts:
Currently, TemplateErrorReason is not being exported, while RenderErrorReason is.
This PR fixes this inconsistency by exporting TemplateErrorReason.


sunng87:
Good catch! Thank you!

PR #596, opened 2023-08-02
Added dev mode and extension filter for RustEmbed

qxuken:
Hi! 

Thanks for the library! 
I have a problem while working with RustEmbed struct. I want a live reload, but now it is implemented only for file register, so i used workaround by iterating struct and register using filenames from it, but i  would like to have it implemented by [handlebars-rust](https://github.com/sunng87/handlebars-rust)

This PR adds dev mode for RustEmbed,  [_which if built not in release mode, reads from the file system on each call_](https://github.com/pyrossh/rust-embed#getfile_path-str---optionrust_embedembeddedfile).
Also added method which names templates and filters files like register_templates_directory does.

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -30,6 +30,8 @@ use rhai::Engine;
 use crate::helpers::scripting::ScriptHelper;
 
 #[cfg(feature = "rust-embed")]
+use crate::sources::LazySource;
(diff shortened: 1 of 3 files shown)


sunng87:
Awesome! This looks fantastic!

PR #601, opened 2023-08-27
Add hidden flag to control when hidden files are ignored

yhakbar:
Would you be willing to accept this adjustment to allow for an additional `hidden` parameter to be passed into `register_templates_directory` to control whether hidden files are ignored?

This would allow for template directories that contained hidden files which should be registered.

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -850,7 +850,8 @@ mod test {
             let mut file4: File = File::create(&file4_path).unwrap();
             writeln!(file4, "<h1>Hallo {{world}}!</h1>").unwrap();
 
-            r.register_templates_directory(".hbs", dir.path(), false).unwrap();
+            r.register_templates_directory(".hbs", dir.path(), false)
+                .unwrap();
 
             assert_eq!(r.templates.len(), 3);
             assert_eq!(r.templates.contains_key("t1"), true);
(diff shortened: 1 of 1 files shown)


src/registry.rs
```diff
@@ -295,6 +296,7 @@ impl<'reg> Registry<'reg> {
         &mut self,
         tpl_extension: &str,
         dir_path: P,
+        hidden: bool,
```
sunng87, line 299: Can we make a `DirectorySourceOptions` struct other potential options?

yhakbar: I can do that, @sunng87 ! 

Do you want any options available in the struct other than `hidden` right now?

I don't know if you ever want temporary files to be pulled in when working with a template directory, but they're checked in the same place, so I can add a `temporary` option too if you want that.

yhakbar: My most recent commit is my best guess as to how this should be handled.

sunng87:
Thank you very much @yhakbar . The patch is almost LGTM. There is a minor lint issue to resolve.

yhakbar:
> Thank you very much @yhakbar . The patch is almost LGTM. There is a minor lint issue to resolve.

Whoops! Should be good now.

sunng87:
rustfmt is broken again, you can fix it and verify from your local code base

yhakbar:
> rustfmt is broken again, you can fix it and verify from your local code base

Formatted.

sunng87:
@yhakbar Thank you!

PR #610, opened 2023-09-24
Re-export JsonTruthy

Dosenpfand:
Or is there a specific reason why this is not done?


Dosenpfand:
@sunng87, thanks for merging this!

> @Dosenpfand do you have a use-case for JsonTruthy? when implementing your own helper?

Yes, exactly. I am using [dotter](https://github.com/SuperCuber/dotter) which uses `handlebars-rust` for with strict mode enabled.

Even with strict mode enabled it would still useful i my opinion, to have a non-strict version of the `and` and `or`, available. Would you accept a PR for that? Otherwise, I would try to get it merged into `dotter` itself.

PR #619, opened 2023-11-06
Support for case conversion helpers.

gabhijit:
Helpers for converting a text string to `camelCase`, `kebab-case`, `snake_case` etc. Supports case conversions from the `heck` crate. This is an opt-in feature which can be enabled using `string_helpers` feature.

### DIFF
--- a/src/helpers/string_helpers/mod.rs
+++ b/src/helpers/string_helpers/mod.rs
@@ -1,0 +1,112 @@
+//! A Set of string utilities often required during code generation.
+//!
+//! See also: [`heck`](https://docs.rs/heck/latest/heck)
+//!
+//! `lowerCamelCase`: Convert a string to lowerCamelCase
+//! `upperCamelCase`: Convert a string to UpperCamelCase
+//! `snakeCase`: Convert a string to snake_case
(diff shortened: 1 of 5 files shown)


src/helpers/string_helpers/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+macro_rules! define_case_helper {
+    ($helper_fn_name: ident, $heck_fn_name:ident) => {
+        pub(crate) fn $helper_fn_name(
+            h: &crate::render::Helper<'_>,
+            _: &crate::Handlebars<'_>,
+            _: &crate::context::Context,
+            _rc: &mut crate::render::RenderContext<'_, '_>,
+            out: &mut dyn crate::output::Output,
+        ) -> crate::helpers::HelperResult {
+            let param = h.param(0).and_then(|v| v.value().as_str()).unwrap_or("");
```
sunng87, line 28: Perhaps we can throw an error for non-string input here?

The error system is slightly different between 4.x and upcoming 5.x so I can do it for you.

sunng87 (approved):

sunng87:
LGTM. I will be cherry-picking this to 4.x branch and update error handling.

PR #630, opened 2024-01-18
refactor: turn off pub access of chain in HelperTemplate

sunng87:
Because `chain` is not used during rendering process, here I turned off pub access of this field for better compatibility.

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -136,7 +136,7 @@ pub struct HelperTemplate {
     pub template: Option<Template>,
     pub inverse: Option<Template>,
     pub block: bool,
-    pub chain: bool,
+    chain: bool,
 }
 
 impl HelperTemplate {


src/template.rs
```diff
@@ -136,7 +136,7 @@ pub struct HelperTemplate {
     pub template: Option<Template>,
     pub inverse: Option<Template>,
     pub block: bool,
-    pub chain: bool,
+    chain: bool,
```
epage, line 139: Wasn't this a breaking change?

Before: https://docs.rs/handlebars/5.1.0/handlebars/template/struct.HelperTemplate.html
After: https://docs.rs/handlebars/latest/handlebars/template/struct.HelperTemplate.html

Both removes access to a field and makes the struct no longer directly initializable.  If the latter is an expected property, this should have `#[non_exhaustive]` (but that too is a breaking change)

sunng87: I see. I didn't realized initialization issue with introducing a private field. Will revert this and yank 5.1.1 for new solution.

epage: This isn't just about initialization but also that a release already exists with `pub chain`

sunng87: This field is introduced in 5.1.0 and not used in any API. The impact should be minimised.

PR #650, opened 2024-06-16
Fix indent after if

cmrschwarz:
Expands on #646 to fix indentation for `{{#if}}` blocks.

### Example Input
```
{{#*inline "partial"}}
<div>
    {{#if foo}}
    foobar
    {{/if}}
</div>
{{/inline}}

<div>
    {{>partial}}
</div>
```
### Output Before
```
<div>
    <div>
    foobar
</div>
</div>
```

### Output Now
```
<div>
    <div>
        foobar
    </div>
</div>
```

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -634,6 +634,11 @@ impl Template {
         // this option is marked as true when standalone statement is detected
         // then the leading whitespaces and newline of next rawstring will be trimed
         let mut trim_line_required = false;
+
+        // for standalone helper block templates, we want the first
+        // raw string to be indented.
+        let mut indent_template_begin = false;
+
         let mut prev_rule = None;
 
         let parser_queue = HandlebarsParser::parse(Rule::handlebars, source).map_err(|e| {
@@ -720,10 +725,24 @@ impl Template {
 
                         let t = template_stack.front_mut().unwrap();
 
-                        // If this text element is following a standalone partial, then
-                        // we trim the whitespace between. But we still want the following text
-                        // to be indented correctly, so we insert the special `Indent` element.
-                        if trim_line_required && prev_rule == Some(Rule::partial_expression) {
+                        let mut emit_indent = false;
+
+                        if trim_line_required {
+                            // If this text element is following a standalone partial, then
+                            // we trim the whitespace between. But we still want the following text
+                            // to be indented correctly, so we insert the special `Indent` element.
+                            emit_indent |= prev_rule == Some(Rule::partial_expression);
+
+                            // if the previous line was the end of a helper block, we ate the whitespace
+                            // so we need to indent here too
+                            emit_indent |= prev_rule == Some(Rule::helper_block_end);
+                        }
+
+                        // if a helper template stars with an inline string, it will
+                        // need to be indented accordingly
+                        emit_indent |= prev_rule == Some(Rule::template) && indent_template_begin;
+
+                        if emit_indent {
                             t.push_element(TemplateElement::Indent, line_no, col_no);
                         }
 
@@ -775,6 +794,8 @@ impl Template {
 
                         let t = template_stack.front_mut().unwrap();
                         t.mapping.push(TemplateMapping(line_no, col_no));
+                        indent_template_begin =
+                            rule == Rule::helper_block_start && trim_line_required;
                     }
                     Rule::invert_tag | Rule::invert_chain_tag => {
                         // hack: invert_tag structure is similar to ExpressionSpec, so I
--- a/src/template.rs
+++ b/src/template.rs
@@ -714,6 +714,13 @@ impl Template {
                 match rule {
                     Rule::template => {
                         template_stack.push_front(Template::new());
+                        if indent_template_begin {
+                            template_stack.front_mut().unwrap().push_element(
+                                TemplateElement::Indent,
+                                line_no,
+                                col_no,
+                            );
+                        }
                     }
                     Rule::raw_text => {
                         // leading space fix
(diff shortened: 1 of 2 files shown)


cmrschwarz:
Hm, this is a little overzealous now if blocks don't contain direct text at all.
Do you have an idea for a better criterion on when to insert the leading indent?

## Breaking Counterexample
```
{{#*inline "nested_partial"}}
<div>
    foobar
</div>
{{/inline}}
{{#*inline "partial"}}
<div>
    {{#if foo}}
    {{#if foo}}
    {{> nested_partial}}
    {{/if}}
    {{/if}}
</div>
{{/inline}}
<div>
    {{>partial}}
</div>
```

## Expected Output 
```
<div>
    <div>
        <div>
            foobar
        </div>
    </div>
</div>
```
## Actual Output
```
<div>
    <div>
            <div>
            foobar
        </div>
    </div>
</div>
```

sunng87:
Just had a quick look at the template of nested if case

```rust
[src/registry.rs:789:9] &tpl = Template {
    name: None,
    elements: [
        RawString(
            "\n",
        ),
        DecoratorBlock(
            DecoratorTemplate {
                name: Name(
                    "inline",
                ),
                params: [
                    Literal(
                        String("nested_partial"),
                    ),
                ],
                hash: {},
                template: Some(
                    Template {
                        name: None,
                        elements: [
                            RawString(
                                "<div>\n    foobar\n</div>\n",
                            ),
                        ],
                        mapping: [
                            TemplateMapping(
                                3,
                                1,
                            ),
                        ],
                    },
                ),
                indent: None,
            },
        ),
        RawString(
            "",
        ),
        DecoratorBlock(
            DecoratorTemplate {
                name: Name(
                    "inline",
                ),
                params: [
                    Literal(
                        String("partial"),
                    ),
                ],
                hash: {},
                template: Some(
                    Template {
                        name: None,
                        elements: [
                            RawString(
                                "<div>\n",
                            ),
                            HelperBlock(
                                HelperTemplate {
                                    name: Name(
                                        "if",
                                    ),
                                    params: [
                                        Path(
                                            Relative(
                                                (
                                                    [
                                                        Named(
                                                            "foo",
                                                        ),
                                                    ],
                                                    "foo",
                                                ),
                                            ),
                                        ),
                                    ],
                                    hash: {},
                                    block_param: None,
                                    template: Some(
                                        Template {
                                            name: None,
                                            elements: [
                                                Indent,
                                                RawString(
                                                    "",
                                                ),
                                                HelperBlock(
                                                    HelperTemplate {
                                                        name: Name(
                                                            "if",
                                                        ),
                                                        params: [
                                                            Path(
                                                                Relative(
                                                                    (
                                                                        [
                                                                            Named(
                                                                                "foo",
                                                                            ),
                                                                        ],
                                                                        "foo",
                                                                    ),
                                                                ),
                                                            ),
                                                        ],
                                                        hash: {},
                                                        block_param: None,
                                                        template: Some(
                                                            Template {
                                                                name: None,
                                                                elements: [
                                                                    Indent,
                                                                    RawString(
                                                                        "    ",
                                                                    ),
                                                                    PartialExpression(
                                                                        DecoratorTemplate {
                                                                            name: Name(
                                                                                "nested_partial",
                                                                            ),
                                                                            params: [],
                                                                            hash: {},
                                                                            template: None,
                                                                            indent: Some(
                                                                                "    ",
                                                                            ),
                                                                        },
                                                                    ),
                                                                    RawString(
                                                                        "",
                                                                    ),
                                                                ],
                                                                mapping: [
                                                                    TemplateMapping(
                                                                        11,
                                                                        5,
                                                                    ),
                                                                    TemplateMapping(
                                                                        11,
                                                                        5,
                                                                    ),
                                                                    TemplateMapping(
                                                                        11,
                                                                        5,
                                                                    ),
                                                                    TemplateMapping(
                                                                        12,
                                                                        5,
                                                                    ),
                                                                ],
                                                            },
                                                        ),
                                                        inverse: None,
                                                        block: true,
                                                        chain: false,
                                                    },
                                                ),
                                                RawString(
                                                    "",
                                                ),
                                            ],
                                            mapping: [
                                                TemplateMapping(
                                                    10,
                                                    5,
                                                ),
                                                TemplateMapping(
                                                    10,
                                                    5,
                                                ),
                                                TemplateMapping(
                                                    10,
                                                    5,
                                                ),
                                                TemplateMapping(
                                                    13,
                                                    5,
                                                ),
                                            ],
                                        },
                                    ),
                                    inverse: None,
                                    block: true,
                                    chain: false,
                                },
                            ),
                            Indent,
                            RawString(
                                "</div>\n",
                            ),
                        ],
                        mapping: [
                            TemplateMapping(
                                8,
                                1,
                            ),
                            TemplateMapping(
                                9,
                                5,
                            ),
                            TemplateMapping(
                                14,
                                1,
                            ),
                            TemplateMapping(
                                14,
                                1,
                            ),
                        ],
                    },
                ),
                indent: None,
            },
        ),
        RawString(
            "<div>\n    ",
        ),
        PartialExpression(
            DecoratorTemplate {
                name: Name(
                    "partial",
                ),
                params: [],
                hash: {},
                template: None,
                indent: Some(
                    "    ",
                ),
            },
        ),
        Indent,
        RawString(
            "</div>\n",
        ),
    ],
    mapping: [
        TemplateMapping(
            1,
            1,
        ),
        TemplateMapping(
            2,
            1,
        ),
        TemplateMapping(
            7,
            1,
        ),
        TemplateMapping(
            7,
            1,
        ),
        TemplateMapping(
            16,
            1,
        ),
        TemplateMapping(
            17,
            5,
        ),
        TemplateMapping(
            18,
            1,
        ),
        TemplateMapping(
            18,
            1,
        ),
    ],
}
```

Perhaps we want to avoid the `Indent` of first `if` ?

cmrschwarz:
Yep, that's the issue. It's what I meant when I said 
> this is a little overzealous now if blocks don't contain direct text at all.

The first {{if}} shouldn't indent, but the second one should, despite both containing a single nested block.
So we need to come up with some rule that we can put in code that leads to the correct behavior.

The correct solution is probably to just make the templates/partials itself start with an `Indent` element,
but I think that would then indent one level too far. Maybe we could give the Indent element an option to
use the parent indent instead of the one of the current block? Though that does not seem like it would win any beauty prizes either. 


Do you have a better Idea? Otherwise I think I could get to trying this approach some time this week.

cmrschwarz:
Superseded by #654

PR #651, opened 2024-06-16
Fix Nested path lookup bug

cmrschwarz:
Fixes a simple bug in path lookup.

## Testcase 
(input: `{"foo": [42], "bar": [1,2,3]}`)
```
let input = r#"
{{#each bar as |x|}}
{{#each ../foo as |y|}}
{{../../foo.0}}: {{../x}}
{{/each}}
{{/each}}
"#;
```
## Output Before
```
42: 
42: 
42: 
```
## Output After
```
42: 1
42: 2
42: 3
```

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -37,7 +37,7 @@ fn parse_json_visitor<'a>(
     block_contexts: &'a VecDeque<BlockContext<'_>>,
     always_for_absolute_path: bool,
 ) -> ResolvedPath<'a> {
-    let mut path_context_depth: i64 = 0;
+    let mut path_context_depth: usize = 0;
     let mut with_block_param = None;
     let mut from_root = false;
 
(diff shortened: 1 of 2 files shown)


sunng87:
LGTM. Thank you!

I'm curious that did you find this in your scenario or using some test suites?

cmrschwarz:
> LGTM. Thank you!
> 
> I'm curious that did you find this in your scenario or using some test suites?

I'm working on a real usecase were I'm creating an interactive debug log/display for a rather complex data processing tool.
All the fixes that I'm PR'ing are just stuff that I run across while developing that tool.

sunng87:
Thank you! I haven't got enough time to look into the if indent fix. Any issue with indent requires me to switch my context and deep dive into it :)

I will take a look in this weekend.

Issue #611, opened 2023-09-27
Extra whitespace added to `each` when the `/each` isn't followed by `\n`


jjcomer:
We noticed this issues when upgrading to `4.x`. If an `/each` is not followed by a `\n` the rendered template will contain extra whitespace. This throws off formats which expect consistent whitespace.

Here is a repro:

```rust
#[test]
    fn no_newline_for_each() {
        let reg = Registry::new();

        let tpl = r#"<ul>
  {{#each a}}
    {{!-- comment --}}
    <li>{{this}}</li>
  {{/each}}"#;
        assert_eq!(
            r#"<ul>
    <li>0</li>
    <li>1</li>
    <li>2</li>"#,
            reg.render_template(tpl, &json!({"a": [0, 1, 2]})).unwrap()
        );
    }
```

Using the template:

```
r#"<ul>
  {{#each a}}
    {{!-- comment --}}
    <li>{{this}}</li>
  {{/each}}"#
```
Results in:

```
---- helpers::helper_each::test::no_newline_for_each stdout ----
thread 'helpers::helper_each::test::no_newline_for_each' panicked at 'assertion failed: `(left == right)`
  left: `"<ul>\n    <li>0</li>\n    <li>1</li>\n    <li>2</li>"`,
 right: `"<ul>\n    <li>0</li>\n      <li>1</li>\n      <li>2</li>\n  "`', src/helpers/helper_each.rs:620:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Using the template:

```
r#"<ul>
  {{#each a}}
    {{!-- comment --}}
    <li>{{this}}</li>
  {{/each}}aa"#
```
Results in:
```
---- helpers::helper_each::test::no_newline_for_each stdout ----
thread 'helpers::helper_each::test::no_newline_for_each' panicked at 'assertion failed: `(left == right)`
  left: `"<ul>\n    <li>0</li>\n    <li>1</li>\n    <li>2</li>"`,
 right: `"<ul>\n    <li>0</li>\n      <li>1</li>\n      <li>2</li>\n  aa"`', src/helpers/helper_each.rs:620:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
Using the template:
```
r#"<ul>
  {{#each a}}
    {{!-- comment --}}
    <li>{{this}}</li>
  {{/each}}a
"#;
```
Results in:
```
-- helpers::helper_each::test::no_newline_for_each stdout ----
[...]

PR #654, opened 2024-06-30
Fix indent v2

cmrschwarz:
This is my second attempt at fixing the more involved indenting issues first brought up in #650 .

It turns out that the `Indent` node is simply not powerful enough to handle all cases :(.

We will unfortunately have to dynamically track whether a partial ended on a newline or not.
This PR does exactly that.

Here are some of the more difficult edge cases that this solves:

### Input
```
{{#*inline "empty_partial"}}{{/inline}}
<div>
    {{> empty_partial}}
</div>
```

### Output
```
<div>
</div>
``` 
---

### Input
```
{{#*inline "partial"}}foo{{/inline}}
<div>
    {{> partial}}
</div>
```

### Output 
```
<div>
    foo</div>
```
(yes, this is indeed correct according to [handlebars js](https://handlebarsjs.com/playground.html#format=1&currentExample=%7B%22template%22%3A%22%7B%7B%23*inline%20%5C%22partial%5C%22%7D%7Dfoo%7B%7B%2Finline%7D%7D%5Cn%3Cdiv%3E%5Cn%20%20%20%20%7B%7B%3E%20partial%7D%7D%5Cn%3C%2Fdiv%3E%22%2C%22partials%22%3A%5B%5D%2C%22input%22%3A%22%7B%7D%5Cn%22%2C%22output%22%3A%22%5Cn%3Cdiv%3E%5Cn%20%20%20%20foo%3C%2Fdiv%3E%22%2C%22preparationScript%22%3A%22%22%2C%22handlebarsVersion%22%3A%224.7.8%22%7D))

---
### Input
Json Input: `{"a": "line\ndynamic_trailing_newline\n"}`
```
{{#*inline "dynamic_partial"}}{{a}}{{/inline}}
<div>
    {{> dynamic_partial}}
</div>
```
### Output
```
<div>
    line
    dynamic_trailing_newline
</div>
```

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -642,7 +642,6 @@ impl Template {
         // this option is marked as true when standalone statement is detected
         // then the leading whitespaces and newline of next rawstring will be trimed
         let mut trim_line_required = false;
-        let mut prev_rule = None;
 
         let parser_queue = HandlebarsParser::parse(Rule::handlebars, source).map_err(|e| {
             let (line_no, col_no) = match e.line_col {
@@ -727,13 +726,6 @@ impl Template {
                         };
 
                         let t = template_stack.front_mut().unwrap();
-
-                        // If this text element is following a standalone partial, then
-                        // we trim the whitespace between. But we still want the following text
-                        // to be indented correctly, so we insert the special `Indent` element.
-                        if trim_line_required && prev_rule == Some(Rule::partial_expression) {
-                            t.push_element(TemplateElement::Indent, line_no, col_no);
-                        }
 
                         t.push_element(
                             Template::raw_string(
@@ -993,7 +985,6 @@ impl Template {
                 if rule != Rule::template {
                     end_pos = Some(span.end_pos());
                 }
-                prev_rule = Some(rule);
             } else {
                 let prev_end = end_pos.as_ref().map(|e| e.pos()).unwrap_or(0);
                 if prev_end < source.len() {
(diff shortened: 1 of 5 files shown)


src/template.rs
```diff
@@ -143,10 +144,11 @@ pub struct HelperTemplate {
     pub inverse: Option<Template>,
     pub block: bool,
     pub chain: bool,
+    pub indent_before_write: bool,
```
sunng87, line 147: I think we can make this field private, as well as the one for `DecoratorTemplate`

sunng87 (commented):
I have been super busy these days. Really need some time to deep dive into this. But the test cases look good. I think this approach is more promising.

src/template.rs
cmrschwarz: Sure!  I need them in `render_helper`, so I guess `pub(crate)` it is.

cmrschwarz:
I've been using this in practice for a bit more now and found two minor issues with it.
The commits just now fixed those and added regression tests.
Otherwise this has worked nicely, indenting seems just about perfect now.

sunng87 (approved):

sunng87:
This looks good to me. Thank you for the fix! I have been struggling with those indent for quite a while. This approach seems practical and also keeps complexity at a acceptable level.

sunng87:
I just tried the case reported in #611, and it seems we still this corner case to cover.

cmrschwarz:
> I just tried the case reported in #611, and it seems we still this corner case to cover.

 I'm happy to look into that, though I'm quite busy this week so it might have to wait until the weekend.

sunng87:
Thank you and no rush @cmrschwarz . Just take your time.

PR #655, opened 2024-06-30
Allow users to set single block params.

cmrschwarz:
This just adds a simple method: `set_block_param` to the `BlockContext`.
It also makes the members of the (already public) `PathAndJson` struct public
so you can make effective use of it.

The existing `set_block_params` only allows replacing all block parameters.
Because there also is no `get_block_params`, this makes it impossible for a helper to
edit the block parameters effectively.

My usecase was to be able to write a helper that binds a variable to a name.
That helper is so useful that I added it to the examples aswell.

### DIFF
--- a/src/json/value.rs
+++ b/src/json/value.rs
@@ -92,8 +92,7 @@ impl<'rc> PathAndJson<'rc> {
     pub fn try_get_constant_value(&self) -> Option<&'rc Json> {
         match &self.value {
             ScopedJson::Constant(value) => Some(*value),
-            ScopedJson::Context(value, _) => Some(*value),
-            ScopedJson::Derived(_) | ScopedJson::Missing => None,
+            ScopedJson::Context(_, _) | ScopedJson::Derived(_) | ScopedJson::Missing => None,
         }
     }
 
--- a/src/json/value.rs
+++ b/src/json/value.rs
@@ -60,8 +60,8 @@ impl<'reg: 'rc, 'rc> From<Json> for ScopedJson<'rc> {
 ///
 #[derive(Debug, Clone)]
 pub struct PathAndJson<'rc> {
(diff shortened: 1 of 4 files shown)


src/json/value.rs
```diff
@@ -60,8 +60,8 @@ impl<'reg: 'rc, 'rc> From<Json> for ScopedJson<'rc> {
 ///
 #[derive(Debug, Clone)]
 pub struct PathAndJson<'rc> {
```
sunng87, line 62: Do we really need these members to be public? There are methods to access them. If it's needed, please add `#[non_exhaustive]` here

cmrschwarz: The problem is that I need string with a `&'rc` lifetime (because that is the key type of BlockParams).
The current accessor functions do not make that possible, and their lifetime annotations cannot be changed to
accomodate that (The helper cannot borrow the whole param for `'rc`).  I guess it would be possible to add a 
`try_get_constant_value(&self) -> &'rc` accessor to that would return `None` for ScopedJson::Derived.
Would you prefer that? Otherwise I'm happy to add `#[non_exhaustive]`.

src/json/value.rs
```diff
@@ -88,6 +88,15 @@ impl<'rc> PathAndJson<'rc> {
         self.value.as_json()
     }
 
+    /// Returns the value, if it is a constant. Otherwise returns None.
+    pub fn try_get_constant_value(&self) -> Option<&'rc Json> {
+        match &self.value {
+            ScopedJson::Constant(value) => Some(*value),
+            ScopedJson::Context(value, _) => Some(*value),
```
sunng87, line 95: This behaviour doesn't seem to align with comment

sunng87 (approved):

PR #659, opened 2024-07-13
apply clippy fixes

cmrschwarz:
This fixes all 86 warnings that clippy currently generates on this project.
Most of these are in test code and don't really matter.
I still believe this is worth it because it makes the use of clippy viable when working on this project.
Currently you are quite likely to miss any issues you introduce because there's already so many warnings.

If you disagree, feel free to close this.

This was partially autogenerated using `cargo clippy --fix`, but manually reviewed & cleaned up.
Should hopfully be very easy to review, as nothing really happens.


sunng87:
Thank you! Because our tests are not covered by github check. Could you please also add `cargo clippy --all-targets` in the github workflow file to ensure we have tests covered?

sunng87 (approved):
I will merge this and add github check myself. Thank you!

PR #661, opened 2024-07-14
Avoid cloning the Render Context 

### LINKED ISSUE #624: Reuse `RenderContext` after executing partial
For now, because of the following line of code, handlebars doens't share `RenderContext` when rendering partial.

https://github.com/sunng87/handlebars-rust/blob/9d7d5556287e31e4148841a56e0981b64a679fb6/src/partial.rs#L61C9-L61C39

```rust
  // Rust
 let partial = find_partial(rc, r, d, tname)?; 
  
 if let Some(t) = partial { 
     // clone to avoid lifetime issue 
     // FIXME refactor this to avoid 
     let mut local_rc = rc.clone();    // <---  This line can't reuse `RenderContext` after partial
  
     // if tname == PARTIAL_BLOCK 
     let is_partial_block = tname == PARTIAL_BLOCK; 
  
     // add partial block depth there are consecutive partial 
```

Is it possible to fix it and have a universal `RenderContext` when `rendering` ?

cmrschwarz:
Inspired by your [comment](https://github.com/sunng87/handlebars-rust/issues/624#issuecomment-2227239418) in #624 I think I came up with a much better way to fix the problem.

The reason why we couldn't get the correct lifetime is that dev mode dynamically creates new templates during rendering. So I just fixed *that* problem instead by making dev mode load all templates upfront.

That way we now no longer have to clone the RenderContext :tada: .  

Incidentally this also solves the issue that dev mode would reload a template from disk **each time** it hits it,
including in some nested ``{{#each}}``. I think this probably more than outweighs the drawback of having to preload all registered templates. We could optimize this further in the future, though it may even be good enough as is.

Resolves #624.


sunng87:
Awesome! This is great step forward!

I will create a benchmark for partials to measure this.

PR #662, opened 2024-07-17
Eliminate `RenderContextInner`

cmrschwarz:
Hey, I realized that there's one last thing I think we could consider before v6.0.

Now that we no longer clone the `RenderContext` we could get rid of the slightly hacky `RenderContextInner`
and the whole writing through an `Rc` idiom associated with it.

I know I initially said that I don't expect any more breaking changes, 
but I missed that this narrows the lifetime contract on the public `RenderContext` type.
This shouldn't really affect anybody (`'reg: 'rc` is basically assumed everwhere already), 
but as a matter of technicality might be semver incompatible.

This should now really be the last breaking change in the indent patch series though :).

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -33,19 +33,15 @@ const BLOCK_HELPER_MISSING: &str = "blockHelperMissing";
 /// This context stores information of a render and a writer where generated
 /// content is written to.
 ///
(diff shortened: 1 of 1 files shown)


sunng87 (approved):
Sounds good to me!

Issue #672, opened 2024-09-06
number comparison functions only accept integer values


lovasoa:
Hi !

Currently, number comparison functions only accept integer values:

https://github.com/sunng87/handlebars-rust/blob/ecef974fa1c9a445adea85e7cdf6f918e284429c/src/helpers/helper_extras.rs#L8-L11

It would be nice if they could compare any Json value with any other, or at least any number with any other.

initially reported in https://github.com/lovasoa/SQLpage/issues/573

sunng87:
Sounds reasonable. I will look into this soon. Thank you!

sunng87:
released in 6.1.0

PR #673, opened 2024-09-06
implement floating point and string comparisons

lovasoa:
fixes https://github.com/sunng87/handlebars-rust/issues/672


sunng87 (approved):
LGTM

lovasoa:
Thanks! When is the next release, so that I can synchronize it with SQLPage?

sunng87:
It has been released in 6.1.0. Just update it and verify.

PR #687, opened 2024-12-20
make "and" and "or" accept any number of parameters

lovasoa:
Writing ifs with multiple conditions used to be very painful, and not very performant

One had to write {{#if (and (cond1 (and cond2 cond3)))}}

One can now write {{#if (and cond1 cond2 cond3)}}


sunng87 (approved):
Thank you!

PR #695, opened 2025-02-25
fix: partial context with literal parameters

### LINKED ISSUE #643: Not work partial context
Incorrect behavior of partial context.
template.hbs
```hbs
{{> partial "hello world"}}
```
partial.hbs
```hbs
{{ this }}
```
result [object] but expected hello world
This also doesn't work if you pass any variable from a shared context to a partial, then the partial will have the same context as the calling template.
handlebars version 5.1.2

sunng87:
This patch add support for literal parameters on partial.

Fixes #643

### DIFF
--- a/src/partial.rs
+++ b/src/partial.rs
@@ -745,5 +745,14 @@ outer third line",
         hbs.register_template_string("t2", t2).unwrap();
 
         assert_eq!("hello world", hbs.render("t2", &()).unwrap());
+
+        let t1 = "{{a}}";
+        let t2 = "{{> t1 \"hello world\" a=1}}";
+
+        let mut hbs = Registry::new();
+        hbs.register_template_string("t1", t1).unwrap();
+        hbs.register_template_string("t2", t2).unwrap();
+
+        assert_eq!("1", hbs.render("t2", &()).unwrap());
     }
 }
--- a/src/partial.rs
+++ b/src/partial.rs
@@ -76,14 +76,21 @@ pub fn expand_partial<'reg: 'rc, 'rc>(
     let mut block_created = false;
 
     // create context if param given
-    // TODO: add support for constants
-    if let Some(base_path) = d.param(0).and_then(|p| p.context_path()) {
-        // path given, update base_path
-        let mut block_inner = BlockContext::new();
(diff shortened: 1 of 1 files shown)


mkantor:
I'm not sure if this is within scope, but it seems that even with this fix, hash params in `this` are not iterated by `#each`. Here's how I tested that:

```diff
--- a/src/partial.rs
+++ b/src/partial.rs
@@ -754,5 +754,14 @@ outer third line",
         hbs.register_template_string("t2", t2).unwrap();
 
         assert_eq!("1", hbs.render("t2", &()).unwrap());
+
+        let t1 = "{{#each this}}{{@key}}:{{this}},{{/each}}";
+        let t2 = "{{> t1 a=1}}";
+
+        let mut hbs = Registry::new();
+        hbs.register_template_string("t1", t1).unwrap();
+        hbs.register_template_string("t2", t2).unwrap();
+
+        assert_eq!("a:1,", hbs.render("t2", &()).unwrap());
     }
 }
```

It turns out this is a recent regression—it works in handlebars-rust v6.3.0 but not v6.3.1 (it also [works in handlebars.js](https://handlebarsjs.com/playground.html#format=1&currentExample=%7B%22template%22%3A%22%7B%7B%3Epartial%20a%3D1%7D%7D%22%2C%22partials%22%3A%5B%7B%22name%22%3A%22partial%22%2C%22content%22%3A%22%7B%7B%23each%20this%7D%7D%7B%7B%40key%7D%7D%3A%7B%7Bthis%7D%7D%2C%7B%7B%2Feach%7D%7D%22%7D%5D%2C%22input%22%3A%22%7B%7D%5Cn%22%2C%22output%22%3A%22a%3A1%2C%22%2C%22preparationScript%22%3A%22%22%2C%22handlebarsVersion%22%3A%224.7.8%22%7D)). Perhaps it was a side effect of #694?

Please let me know if you'd like me to file a separate issue about this.

sunng87:
@mkantor Thank you for reporting! In #694 I changed hash implementation from merging into context to block param. It seems we still need to merge the hash into the partial context for `this`. 

But this is another issue can be address in separated PR. I will merge this and work on the issue when I have ideas.

Issue #696, opened 2025-02-25
Merging partial context with hash, regression of partial context refactor


sunng87:
I'm not sure if this is within the scope of this pull request, but it seems that even with this fix, hash params in `this` are not iterated by `#each`. Here's how I tested that:

```diff
--- a/src/partial.rs
+++ b/src/partial.rs
@@ -754,5 +754,14 @@ outer third line",
         hbs.register_template_string("t2", t2).unwrap();
 
         assert_eq!("1", hbs.render("t2", &()).unwrap());
+
+        let t1 = "{{#each this}}{{@key}}:{{this}},{{/each}}";
+        let t2 = "{{> t1 a=1}}";
+
+        let mut hbs = Registry::new();
+        hbs.register_template_string("t1", t1).unwrap();
+        hbs.register_template_string("t2", t2).unwrap();
+
+        assert_eq!("a:1,", hbs.render("t2", &()).unwrap());
     }
 }
```

(Same problem occurs when using block param syntax, e.g. `{{#each this as |value key|}}{{key}}:{{value}},{{/each}}`).

It turns out this is a recent regression—it works in handlebars-rust v6.3.0 but not v6.3.1 (it also [works in handlebars.js](https://handlebarsjs.com/playground.html#format=1&currentExample=%7B%22template%22%3A%22%7B%7B%3Epartial%20a%3D1%7D%7D%22%2C%22partials%22%3A%5B%7B%22name%22%3A%22partial%22%2C%22content%22%3A%22%7B%7B%23each%20this%7D%7D%7B%7B%40key%7D%7D%3A%7B%7Bthis%7D%7D%2C%7B%7B%2Feach%7D%7D%22%7D%5D%2C%22input%22%3A%22%7B%7D%5Cn%22%2C%22output%22%3A%22a%3A1%2C%22%2C%22preparationScript%22%3A%22%22%2C%22handlebarsVersion%22%3A%224.7.8%22%7D)). Perhaps it was a side effect of #694?

Please let me know if you'd like me to file a separate issue about this.

_Originally posted by @mkantor in https://github.com/sunng87/handlebars-rust/issues/695#issuecomment-2681899919_

PR #699, opened 2025-03-15
feat: correct implementation of partial block

### LINKED ISSUE #698: 6.3.1 generates incorrect output
I have a custom static site generator at https://github.com/hrydgard/ppsspp-site. It breaks really badly with 6.3.1, while it worked perfectly in 6.3.0.

To reproduce, just clone the linked repo and type cargo run, then go to localhost:3000.

You'll see that the site headers are all a single repeated string instead of the correct strings - but not on all pages! This points to it being some hashing issue, which makes it likely related to #696 .

sunng87:
Fixes #698 #696 

This patch finally reimplements partial's block context correctly. The key is to remove all previous blocks and use a separated one for partial.

It corrects behavior of partial parameter, hash and merge_json:
- if parameters is provided, either as path or literal, use it as base value; otherwise use `this` as base_value
- if hash is provided, always merge it with base value:
  - if base value is associative type: object, array, string, merge them as map
  - if base value is non-associative type, return hash
  otherwise, use base value directly

When partial is rendered, reset blocks.

Tests added.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -2,6 +2,7 @@ use std::collections::{HashMap, VecDeque};
 
 use serde::Serialize;
 use serde_json::value::{to_value, Value as Json};
+use serde_json::Map;
 
 use crate::block::{BlockContext, BlockParamHolder};
 use crate::error::{RenderError, RenderErrorReason};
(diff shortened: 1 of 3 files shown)


hrydgard:
Thank you!

sunng87:
Just trying to validate on your project.

sunng87:
@hrydgard it should work now. I will look into another partial issue tomorrow and cut a release soon.

hrydgard:
Finally got around to updating. It works fine, thank you!

PR #707, opened 2025-04-17
docs: remove incorrect parameters from docstring

### LINKED ISSUE #705: Docstring inconsistency in function `register_templates_directory`
Docstring references parameters `tpl_extension` and `dir_path`, while function declaration references `dir_path` and `options`.

Looks like #601 introduced the change, but these two doc lines were overlooked.

https://github.com/sunng87/handlebars-rust/blob/7d666fb7cf117a951f44abb2a5a40029bd627ef7/src/registry.rs#L326-L348

Other functions in the file don't seem to have this type of annotation. Should I send a PR just deleting these lines to avoid confusion?

https://github.com/sunng87/handlebars-rust/blob/7d666fb7cf117a951f44abb2a5a40029bd627ef7/src/registry.rs#L328-L330

vsbits:
Closes #705

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -325,9 +325,6 @@ impl<'reg> Registry<'reg> {
 
     /// Register templates from a directory
     ///
-    /// * `tpl_extension`: the template file extension
-    /// * `dir_path`: the path of directory
-    ///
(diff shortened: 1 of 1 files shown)


sunng87:
Thank you!

PR #710, opened 2025-05-13
expose DirectorySourceOptionsBuilder & added doc example

markwellis:
DirectorySourceOptions is `#[non_exhaustive]` which means it cannot be created directly by users of the crate.

DirectorySourceOptions is `#[derive(Builder)]` but the resulting DirectorySourceOptionsBuilder is not exposed to users of the crate.

Because of this only `DirectorySourceOptions::default()` can be used and options cannot be set.

Exposing DirectorySourceOptionsBuilder allows users of this crate to build DirectorySourceOptions.


sunng87:
@markwellis That's a good idea. Please check the CI issue and I will merge this once it's fixed. Thank you!

PR #713, opened 2025-05-25
feat: Added dynamic partial blocks by adding support for empty block closing tags

henriquecolini:
This PR implements a feature that has been historically lacking from the original Handlebars: dynamic partial blocks. https://github.com/handlebars-lang/handlebars-parser/issues/9

Handlebars syntax allows users to create both dynamic partials:
```handlebars
{{> (snake_case 'MyPartial') }}
```

And partial blocks:
```handlebars
{{#> my_partial }}
    Hello
{{/my_partial}}
```

In theory, one should be able to use both features at the same time since there's no conflict whatsoever betwen them. In practice though, this turns out to be impossible due to a funny oversight in the syntax:
```handlebars
{{> (snake_case 'MyPartial') }}
    Hello
{{/ ??? }}
what goes here?
```

I have fixed this issue by making the explicit name in the closing tag optional, which I consider the most obvious change to the syntax. So, this is now possible:
```handlebars
{{> (snake_case 'MyPartial') }}
    Hello, this works now!
{{/}}
```

And as a bonus, it's also possible in general:
```
{{> my_partial }}
    {{#*inline foo }}
        some inline stuff
    {{/}}
{{/}}
```

### DIFF
--- a/src/grammar.pest
+++ b/src/grammar.pest
@@ -88,14 +88,14 @@ helper_block = _{ helper_block_start ~ template ~
 
 decorator_block_start = { "{{" ~ leading_tilde_to_omit_whitespace? ~ "#" ~ "*"
                         ~ exp_line ~ trailing_tilde_to_omit_whitespace? ~ "}}" }
-decorator_block_end = { "{{" ~ leading_tilde_to_omit_whitespace? ~ "/" ~ identifier ~
+decorator_block_end = { "{{" ~ leading_tilde_to_omit_whitespace? ~ "/" ~ identifier? ~
                         trailing_tilde_to_omit_whitespace? ~ "}}" }
 decorator_block = _{ decorator_block_start ~ template ~
                      decorator_block_end }
 
 partial_block_start = { "{{" ~ leading_tilde_to_omit_whitespace? ~ "#" ~ ">"
                         ~ partial_exp_line ~ trailing_tilde_to_omit_whitespace? ~ "}}" }
-partial_block_end = { "{{" ~ leading_tilde_to_omit_whitespace? ~ "/" ~ partial_identifier ~
+partial_block_end = { "{{" ~ leading_tilde_to_omit_whitespace? ~ "/" ~ partial_identifier? ~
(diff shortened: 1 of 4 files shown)


src/grammar.pest
```diff
@@ -81,27 +83,27 @@ invert_chain_tag = { !escape ~ "{{" ~ leading_tilde_to_omit_whitespace? ~ invert
                      ~ exp_line ~ trailing_tilde_to_omit_whitespace? ~ "}}" }
 helper_block_start = { "{{" ~ leading_tilde_to_omit_whitespace? ~ "#" ~ exp_line ~
                      trailing_tilde_to_omit_whitespace? ~ "}}" }
-helper_block_end = { "{{" ~ leading_tilde_to_omit_whitespace? ~ "/" ~ identifier ~
+helper_block_end = { "{{" ~ leading_tilde_to_omit_whitespace? ~ "/" ~ opt_identifier ~
```
sunng87, line 86: Can we just do identifier? here?

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "handlebars"
-version = "6.3.2"
+version = "6.3.3"
```
sunng87, line 3: We don't change version number in feature patch.

sunng87 (commented):
This looks mostly good to me. Just add two small comments

src/grammar.pest
henriquecolini: If we use identifier in that line, that disables the support of empty closing "tags" for helper blocks. I'm not sure what is the use case for that would be, but I do like to keep it consistent across all the blocks of the language.
```handlebars
{{to_uppercase}}
my string
{{/}}
```

Cargo.toml
henriquecolini: Yeah, that was a mistake on my part. Should I make a commit reverting that change?

sunng87, line 3:
```suggestion
version = "6.3.2"
```

sunng87:
@henriquecolini I just reverted the version change by myself. Thank you for the patch!

PR #716, opened 2025-06-30
Add a helper for markdown

smclmnt:
this adds a feature `markdown_helpers` which provides a helper for handling markdown text which is enabled by default when the feature is enabled.

The feature `markdown_fmt_default` can be used toggle to the CommonMark specification instead of the GitHub one.


sunng87:
@smclmnt Thank you for the patch. In order to keep the core library simple and compact, I will not include this in handlebars-rust library. However, thanks to the extensible design, you can build a library , say handlebars-markdown-helper, for this. Actually I remember there is such a library.

PR #725, opened 2025-11-09
Add Ability to Unregister Helpers

### LINKED ISSUE #724: Add ability to unregister helpers
Hello, 

First of all this is an awesome project, I've been using to add template support to one of my projects!

However I noticed I'm not able to to unregister `helpers` and in addition `script_helpers`. 

I was wondering if there a is a certain reason for this behavior?

The reason why I want to be able to unregister helpers is I want to use `script_helpers` to add the ability to modify values for templates but I don't necessarily want them to add them globally.

I took a quick look at the code for the `registry` module:

https://github.com/sunng87/handlebars-rust/blob/507cbf214e89f2e1e1687ebd2f9bbae04b999ca5/src/registry.rs#L489-L492

It looks like it just adds the boxed `HelperDef` trait to a `HashMap`, so couldn't there be a function like the following that does the opposite?:

```rust
/// Unregister a helper
pub fn unregister_helper(&mut self, name: &str) {
    self.helpers.remove(name);
}
```

I would be happy to create a pull request if this is and acceptable solution.

If this feature doesn't get added I don't mind, as I believe I can just clone the handlebars struct in rendering function of my project then register the helpers so it creates a temporary registration, though I haven't tested this solution yet.

Lepidopteran:
Let me know if I need to make any changes!

Resolves #724

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -927,6 +927,27 @@ mod test {
     static DUMMY_HELPER: DummyHelper = DummyHelper;
 
     #[test]
+    fn test_unregister_helper() {
+        let mut r = Registry::new();
+
+        assert!(r
+            .register_template_string("dumber", "{{#dumb}}\ndummy helper exists\n{{/dumb}}")
+            .is_ok());
+
+        r.register_helper("dumb", Box::new(DUMMY_HELPER));
+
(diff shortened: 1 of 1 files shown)


Lepidopteran:
@sunng87 Unit test has been added!

Issue #729, opened 2025-12-03
Support for recursive lookup?


wez:
I'm evaluating handlebars-rust where it needs to be able to support this sort of recursive lookup scenario as part of supporting some legacy applications with existing templates:

```handlebars
{{#each children}}
   outer={{{outer_value}}}
   inner={{{inner_value}}}
{{/each}}
```

with input data:

```json
{
  "children": [{"inner_value": "foo"}],
  "outer_value": "outer"
}
```

the legacy application runs with handlebars.js running in `compat` mode:

> Handlebars deviates from Mustache slightly in that it does not perform recursive lookup by default. The compile time `compat` flag must be set to enable this functionality. Users should note that there is a performance cost for enabling this flag. The exact cost varies by template, but it's recommended that performance sensitive operations should avoid this mode and instead opt for explicit path references.

so the template renders as:

```
outer=outer
inner=foo
```

handlebars-rust renders it as:

```
outer=
inner=foo
```

Is adding optional support for this kind of recursive lookup something that you'd want in handlebars-rust?

I'm willing to look at working up a PR to support this if you can point me to the roughly the right place!

wez:
I can turn this into a PR if this desirable and seems right:

```diff
diff --git a/src/context.rs b/src/context.rs
index df5ddf6..6181095 100644
--- a/src/context.rs
+++ b/src/context.rs
@@ -197,13 +197,57 @@ impl Context {
         let resolved_visitor = parse_json_visitor(relative_path, block_contexts, true);

         match resolved_visitor {
-            ResolvedPath::AbsolutePath(paths) => {
-                let mut ptr = Some(self.data());
-                for p in &paths {
-                    ptr = get_data(ptr, p)?;
+            ResolvedPath::AbsolutePath(mut paths) => {
+                // Only attempt a recursive resolution if it looks like a simple
+                // named parameter, rather than an explicitly pathed parameter like
+                // "./foo" or "../foo".
+                let allow_recursive = matches!(relative_path, [PathSeg::Named(_)]);
+                if allow_recursive {
+                    // Paths is probably something like: ["children", "0", "name"].
+                    // This block of code will first try to resolve that set of
+                    // paths, but if no value is found, it will then successively
+                    // mutate that list to remove the parent element (the penultimate
+                    // one) and resolving each in turn:
+                    // ["children", "name"]
+                    // ["name"]
+                    while !paths.is_empty() {
+                        let mut ptr = Some(self.data());
[...]

PR #730, opened 2025-12-05
Option to enable recursive resolution of variables

wez:
refs: https://github.com/sunng87/handlebars-rust/issues/729


sunng87 (approved):
LGTM. Thank you @wez

